PR #1, opened 2021-04-26
add Drives Endpoints

micbar:
# Description

This holds the spec for https://github.com/owncloud/ocis/blob/de36293a72dd552ccb37a035926fb2a84b425ad5/docs/adr/0005-api-for-spaces.md

PR https://github.com/owncloud/ocis/pull/1827

### DIFF
--- a/api/openapi-spec/v0.0.yml
+++ b/api/openapi-spec/v0.0.yml
@@ -1,0 +1,933 @@
+openapi: 3.0.1
+info:
+  title: Open Graph API
+  version: v0.0
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+servers:
+  - url: https://graph.owncloud.com/v0.0/
+    description: Core
+paths:
+  /me/drives:
+    get:
+      tags:
+        - me.drive
+      summary: Get drives from me
+      operationId: me.ListDrives
+      parameters:
+        - $ref: '#/components/parameters/top'
+        - $ref: '#/components/parameters/skip'
+        - $ref: '#/components/parameters/search'
+        - $ref: '#/components/parameters/filter'
+        - $ref: '#/components/parameters/count'
+        - name: $select
+          in: query
+          description: Select properties to be returned
+          style: form
+          explode: false
+          schema:
+            uniqueItems: true
+            type: array
+            items:
+              enum:
+                - id
+                - createdBy
+                - createdDateTime
+                - description
+                - eTag
+                - lastModifiedBy
+                - lastModifiedDateTime
+                - name
+                - parentReference
+                - webUrl
+                - driveType
+                - oCDriveStatus
+                - owner
+                - oCCoOwner
+                - quota
+                - createdByUser
+                - lastModifiedByUser
+                - root
+              type: string
+            maxItems: 16
+        - name: $expand
+          in: query
+          description: Expand related entities
+          style: form
+          explode: false
+          schema:
+            uniqueItems: true
+            type: array
+            items:
+              enum:
+                - '*'
+                - root
+              type: string
+            maxItems: 2
+      responses:
+        '200':
+          description: Retrieved navigation property
+          content:
+            application/json:
+              schema:
+                title: Collection of drive
+                type: object
+                properties:
+                  value:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/open.graph.drive'
+                    maxItems: 100
+                  '@odata.nextLink':
+                    type: string
+          links:
+            root:
+              operationId: me.drive.root.GetRoot
+              parameters:
+                drive-id: $request.path.drive-id
+        default:
+          $ref: '#/components/responses/error'
+      x-ms-pageable:
+        nextLinkName: '@odata.nextLink'
+        operationName: listMore
+      x-ms-docs-operation-type: operation
+    post:
+      tags:
+        - me.drive
+      summary: Create new navigation property to drives for me
+      operationId: me.CreateDrives
+      requestBody:
+        description: New navigation property
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/open.graph.drive'
+        required: true
+      responses:
+        '201':
+          description: Created navigation property.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/open.graph.drive'
+        default:
+          $ref: '#/components/responses/error'
+      x-ms-docs-operation-type: operation
+  '/me/drives/{drive-id}':
+    get:
+      tags:
+        - me.drives
+      summary: Get drive by id
+      operationId: me.drives.GetDrive
+      parameters:
+        - name: drive-id
+          in: path
+          description: 'key: id of drive'
+          required: true
+          schema:
+            type: string
+          x-ms-docs-key-type: drive
+        - name: $select
+          in: query
+          description: Select properties to be returned
+          style: form
+          explode: false
+          schema:
+            uniqueItems: true
+            type: array
+            items:
+              enum:
+                - id
+                - createdBy
+                - createdDateTime
+                - description
+                - eTag
+                - lastModifiedBy
+                - lastModifiedDateTime
+                - name
+                - parentReference
+                - webUrl
+                - driveType
+                - oCDriveStatus
+                - owner
+                - oCCoOwner
+                - quota
+                - createdByUser
+                - lastModifiedByUser
+                - root
+              type: string
+            maxItems: 16
+        - name: $expand
+          in: query
+          description: Expand related entities
+          style: form
+          explode: false
+          schema:
+            uniqueItems: true
+            type: array
+            items:
+              enum:
+                - '*'
+                - root
+              type: string
+            maxItems: 2
+      responses:
+        '200':
+          description: Retrieved navigation property
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/open.graph.drive'
+          links:
+            root:
+              operationId: me.drive.root.GetRoot
+              parameters:
+                drive-id: $request.path.drive-id
+        default:
+          $ref: '#/components/responses/error'
+      x-ms-docs-operation-type: operation
+    patch:
+      tags:
+        - me.drives
+      summary: Update the navigation property of a specific drive
+      operationId: me.drives.UpdateDrive
+      parameters:
+        - name: drive-id
+          in: path
+          description: 'key: id of drive'
+          required: true
+          schema:
+            type: string
+          x-ms-docs-key-type: drive
+      requestBody:
+        description: New navigation property values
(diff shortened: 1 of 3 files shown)


api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,992 @@
...
+          anyOf:
+            - $ref: '#/components/schemas/open.graph.folderView'
+          description: A collection of properties defining the recommended view for the folder.
+          nullable: true
+    open.graph.image:
+      title: image
+      type: object
+      properties:
+        height:
+          maximum: 2147483647
```
butonic, line 803: so image sizes will only get bigger http://www.gigamacro.com/worlds-first-terapixel-macro-image/

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,992 @@
...
+          minimum: -2147483648
```
butonic, line 804: why are negative sizes ok, then?

butonic (approved):
I feel a bit tired today. Take the review with a grain of salt. The only thing I really dislike is having to implement all the ordering from the start...

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,992 @@
...
+            schema:
+              $ref: '#/components/schemas/open.graph.drive'
+        required: true
+      responses:
+        '201':
+          description: Created navigation property.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/open.graph.drive'
```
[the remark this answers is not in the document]
micbar: we can technically reference a schema from a remote resource https://swagger.io/docs/specification/using-ref/

our open.graph.drive is already a bit different (oCCoOwner, oCDriveStatus), so we need to maintain it.

api/openapi-spec/v0.0.yml
micbar: IMO this value refers to the maximum length of our datatype for that attribute

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,992 @@
...
+        - me.drive
+      summary: Get drives from me
+      operationId: me.ListDrives
+      parameters:
+        - $ref: '#/components/parameters/top'
+        - $ref: '#/components/parameters/skip'
+        - $ref: '#/components/parameters/search'
+        - $ref: '#/components/parameters/filter'
+        - $ref: '#/components/parameters/count'
+        - name: $orderby
```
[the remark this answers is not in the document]
micbar: can skip it now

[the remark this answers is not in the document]
micbar: removed the sort parameter

api/openapi-spec/v0.0.yml
micbar: seems this is int32

are we limited by the extended attributes?

api/openapi-spec/v0.0.yml
micbar: seems weird.

micbar:
@refs @butonic @kulmann 
according to the spec we should add a response `401` when the user doesn't provide a bearer token.

dragotin (approved):
Only unqualified comments from me, as I am not fluid with the format.

api/openapi-spec/v0.0.yml
micbar: removed it

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,950 @@
...
+                - parentReference
+                - webUrl
+                - driveType
+                - oCDriveStatus
+                - owner
+                - oCCoOwner
+                - quota
+                - createdByUser
+                - lastModifiedByUser
+                - root
```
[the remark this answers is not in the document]
micbar: `createdBy`: Identity of the user, device, or application which created the item. Read-only.
Referencing an identity set

`createdByUser`: Identity of the user who created the item.
Referencing a user object.

Graph API makes a difference there. IMO makes sense.

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,950 @@
...
+              type: string
+            maxItems: 16
```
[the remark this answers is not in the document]
micbar: thanks, corected.

micbar:
@butonic @dragotin IMO this could be merged now. Please review again.

@refs I removed the title atribute completely. I found no other public resource using it. For me it works perfectly without it.

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,970 @@
...
+        - type: object
+          properties:
+            createdBy:
+              anyOf:
+                - $ref: '#/components/schemas/open.graph.identitySet'
+              description: 'Identity of the user, device, or application which created the item. Read-only.'
+              nullable: true
+              readOnly: true
+            createdDateTime:
+              pattern: '^[0-9]{4,}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]([.][0-9]{1,12})?(Z|[+-][0-9][0-9]:[0-9][0-9])$'
```
C0rby, line 401: RFC3339 Allows `T` and `Z` to be lower case.

I'd suggest this changes '^[0-9]{4,}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])**[Tt]**([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]([.][0-9]{1,12})?(**[Zz]**|[+-][0-9][0-9]:[0-9][0-9])$'

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,970 @@
...
+              description: The telephone numbers for the user. Only one number can be set for this property. Returned by default. Read-only for users synced from on-premises directory.
+            city:
+              type: string
+              description: The city in which the user is located. Returned only on $select. Supports $filter.
+              nullable: true
+            companyName:
+              type: string
+              description: The company name which the user is associated. This property can be useful for describing the company that an external user comes from. The maximum length of the company name is 64 characters.Returned only on $select.
+              nullable: true
+            country:
```
C0rby, line 512: Will we be using the ISO 3166 Country Codes here?
If so we could limit the size to 2 or 3 depending on if we use the Alpha-2 or Alpha-3 Codes

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,970 @@
...
+        - $ref: '#/components/schemas/open.graph.directoryObject'
+        - type: object
+          properties:
+            accountEnabled:
+              type: boolean
+              description: 'true if the account is enabled; otherwise, false. This property is required when a user is created. Returned only on $select. Supports $filter.'
+              nullable: true
+            businessPhones:
+              type: array
+              items:
```
C0rby, line 501: It should be possible to limit this value to a size right?
There must be some spec for telephone numbers.

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,970 @@
...
+            officeLocation:
+              type: string
+              description: The office location in the user's place of business. Returned by default.
+              nullable: true
+            postalCode:
+              type: string
+              description: 'The postal code for the user''s postal address. The postal code is specific to the user''s country/region. In the United States of America, this attribute contains the ZIP code. Returned only on $select.'
+              nullable: true
+            preferredLanguage:
+              type: string
```
C0rby, line 571: We should be able to limit the size of this attribute since the description says it should contain ISO 639-1 Codes

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,970 @@
...
+              format: date-time
+              nullable: true
+          description: Represents an Active Directory object. The directoryObject type is the base type for many other directory entity types.
+    open.graph.quota:
+      type: object
+      properties:
+        deleted:
+          type: integer
+          description: 'Total space consumed by files in the recycle bin, in bytes. Read-only.'
+          format: int64
```
C0rby, line 814: Just a remark. int64 will limit us to exabyte. I don't know how long until someone actually has so much storage capabilities but we should keep that in mind.
On the other hand...this could always be changed in a future API version.

P.S.: If the 'e' in eos really stand for exabyte then it might be coming earlier. :b

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,970 @@
...
+      properties:
+        crc32Hash:
+          type: string
+          description: The CRC32 value of the file (if available). Read-only.
+          nullable: true
+        quickXorHash:
+          type: string
+          description: A proprietary hash of the file that can be used to determine if the contents of the file have changed (if available). Read-only.
+          nullable: true
+        sha1Hash:
```
C0rby, line 863: Do we still want to use SHA-1 in new APIs?
SHA-1 is considered broken. https://en.wikipedia.org/wiki/SHA-1

api/openapi-spec/v0.0.yml
```diff
@@ -0,0 +1,970 @@
...
+      properties:
+        upgradeAvailable:
+          type: boolean
+          description: Indicates if there are higher storage quota plans available. Read-only.
+          nullable: true
+    open.graph.hashes:
+      type: object
+      properties:
+        crc32Hash:
+          type: string
```
C0rby, line 856: The size of the hashes can be limited.
8 for crc32 (8 hexadecimal characters -> 32 bytes)
40 for SHA-1 (160 bytes)
64 for SHA-256 (256 bytes)

C0rby (commented):
Some small remarks.

api/openapi-spec/v0.0.yml
micbar: did you miss a group?

micbar: got it

Changed it on all occurrances.

api/openapi-spec/v0.0.yml
micbar: let us limit to 3

api/openapi-spec/v0.0.yml
micbar:
```
pattern: '^[a-z]{2}(-[A-Z]{2})?$'
maxLength: 7
```

api/openapi-spec/v0.0.yml
micbar: I would vote for not restricting that. So unpredictable.

C0rby: Fine by me.

api/openapi-spec/v0.0.yml
C0rby: I think that should work. :+1:

C0rby (approved):

PR #3, opened 2021-05-12
Fix code generation

butonic:
OpenAPI v3.1 [seems to alow $ref siblings](https://stackoverflow.com/questions/33629750/swagger-schema-properties-ignored-when-using-ref-why/41752575#41752575), which allows us to generate actually usable code. So we dropped the anyOf "workaround".

To validate the 3.1 spec the [openapi-generator in #9083 tracks](https://github.com/OpenAPITools/openapi-generator/issues/9083) the [swagger-api/swagger-parser #1531](https://github.com/swagger-api/swagger-parser/issues/1535) implementation for it.


refs (approved):
👍 LGTM, doc generation will probably explode with openapi-generator because we're using a different generator, so I would leave documentation out of CI and instead rely on godocs

micbar:
I copy&pasted the spec into swagger editor, seems fine so far.

micbar:
CI validator is also green.

refs:
Continuing with this code generation library, we'd need to also update: https://github.com/owncloud/open-graph-api-go

PR #8, opened 2021-11-24
Update the specs to match the current implementation

micbar:
# Description

- Cleanup and remove unneeded attributes
- Change the routes to match the official specs

### DIFF
--- a/api/openapi-spec/v0.0.yaml
+++ b/api/openapi-spec/v0.0.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.1
 info:
   title: Open Graph API
-  version: v0.0
+  version: v0.1.0
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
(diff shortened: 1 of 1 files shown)


butonic (approved):
nice!

api/openapi-spec/v0.0.yaml
```diff
@@ -213,11 +187,29 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
+    post:
+      tags:
+        - drives
+      summary: Create a new navigation property of a specific drive
```
refs, line 193: The operation id doesn't match the description. I think it is a copy paste leftover from the previous operation.

refs (commented):
Can't request changes because it is merged, but worth having a look at that description @micbar ^^

api/openapi-spec/v0.0.yaml
micbar: operation-id is `drives.CreateDrive`

PR #16, opened 2021-12-02
Remove the allOf model composition

micbar:
# Description

Remove the model composition and repeat the sub-type patterns by adding them.
This breaks DRY but should make the code generation possible at all.

### DIFF
--- a/api/openapi-spec/v0.0.yaml
+++ b/api/openapi-spec/v0.0.yaml
@@ -306,96 +306,85 @@ components:
     drive:
       description: Storage Space. Read-only.
       readOnly: true
-      allOf:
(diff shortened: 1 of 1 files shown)


TheOneRing:
Don't think so ` error C3646: 'getDrive': unknown override specifier`

micbar:
> Don't think so ` error C3646: 'getDrive': unknown override specifier`

Could you help me to understand the error?

TheOneRing:
> > Don't think so ` error C3646: 'getDrive': unknown override specifier`
> 
> Could you help me to understand the error?

The type is not defined, either due to circular or missing include.

PR #18, opened 2021-12-14
Add user groups

micbar:
# Description

## Add User endpoints

`/users/{User-ID}`
`/users/`

## Add group schema

With a selected set of properties

## Add group endpoints

`/groups/`
`/groups/{Group-ID}`

### DIFF
--- a/api/openapi-spec/v0.0.yaml
+++ b/api/openapi-spec/v0.0.yaml
@@ -543,6 +543,400 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
+  /users:
+    get:
+      tags:
+        - users
+      summary: Get entities from users
+      operationId: ListUsers
+      parameters:
+        - $ref: '#/components/parameters/top'
+        - $ref: '#/components/parameters/skip'
+        - $ref: '#/components/parameters/search'
+        - $ref: '#/components/parameters/filter'
+        - $ref: '#/components/parameters/count'
+        - name: $orderby
+          in: query
+          description: Order items by property values
+          style: form
+          explode: false
+          schema:
+            uniqueItems: true
+            type: array
+            items:
+              enum:
+                - id
+                - id desc
+                - deletedDateTime
+                - deletedDateTime desc
+                - accountEnabled
+                - accountEnabled desc
+                - businessPhones
+                - businessPhones desc
+                - city
+                - city desc
+                - companyName
+                - companyName desc
+                - country
+                - country desc
+                - createdDateTime
+                - createdDateTime desc
+                - department
+                - department desc
+                - displayName
+                - displayName desc
+                - faxNumber
+                - faxNumber desc
+                - givenName
+                - givenName desc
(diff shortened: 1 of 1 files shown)


api/openapi-spec/v0.0.yaml
```diff
@@ -495,6 +1082,36 @@ components:
...
+        onPremisesSyncEnabled:
+          type: boolean
+          description: 'true if this object is synced from an on-premises directory; false if this object was originally synced from an on-premises directory but is no longer synced; null if this object has never been synced from an on-premises directory (default). Read-only. Returned only on $select. Supports $filter (eq, ne, not, in, and eq on null values).'
+          nullable: true
+        onPremisesLastSyncDateTime:
+          pattern: '^[0-9]{4,}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]([.][0-9]{1,12})?(Z|[+-][0-9][0-9]:[0-9][0-9])$'
+          type: string
+          description: 'Indicates the last time at which the object was synced with the on-premises directory; for example: 2013-02-16T03:04:54Z. The Timestamp type represents date and time information using ISO 8601 format and is always in UTC time. For example, midnight UTC on Jan 1, 2014 is 2014-01-01T00:00:00Z. Read-only. Returned only on $select. Supports $filter (eq, ne, not, ge, le, in).'
+          format: date-time
+          nullable: true
```
rhafer, line 1106: What's reason for adding all of these? I doubt we will need any of this anytime soon, if we ever need it.

As already mentioned elsewhere, I somehow think we should be the list of attributes small for now (if only to avoid confusion about what we actually support).

api/openapi-spec/v0.0.yaml
```diff
@@ -843,6 +1464,124 @@ components:
...
+          description: Read-only.
+          readOnly: true
+        # directory object
+        deletedDateTime:
+          pattern: '^[0-9]{4,}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])[Tt]([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]([.][0-9]{1,12})?([Zz]|[+-][0-9][0-9]:[0-9][0-9])$'
+          type: string
+          format: date-time
+          nullable: false
+      description: 'The user (or application) that created the group. NOTE: This is not set if the user is an administrator. Read-only.'
+      nullable: true
```
rhafer, line 1584: The description of `directoryObject` seems wrong. (copy and paste error?)

rhafer (changes requested):
I think we shouldn't be adding all of those attributes that we're likely to never support (but I guess I can live with that)

Please fix the `directorObject` description though.

api/openapi-spec/v0.0.yaml
micbar: yes. Corrected.

rhafer (approved):

refs (approved):

PR #23, opened 2022-01-20
Add endpoints for managing group membership

rhafer:
- [x] POST /groups/{id}/members/$ref (for adding a single member)
 - [x] DELETE graph/v1.0/groups/{id}/members/{id}/$ref (for deleting a single member)
 - [x] PATCH graph/v1.0/groups/{group-id} (for adding multiple members)

### DIFF
--- a/api/openapi-spec/v0.0.yaml
+++ b/api/openapi-spec/v0.0.yaml
@@ -498,6 +498,68 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
+  '/groups/{group-id}/members/$ref':
+    post:
+      tags:
+        - group
+      summary: Add a member to a group
+      operationId: AddMember
+      parameters:
+        - name: group-id
+          in: path
+          description: 'key: id of group'
+          required: true
+          schema:
+            type: string
+          x-ms-docs-key-type: group
+      requestBody:
+        description: Object to be added as member
+        content:
+          application/json:
+            schema:
+              title: Member Reference
+              type: object
+              properties:
+                '@odata.id':
+                  type: string
+        required: true
+      responses:
+        '204':
+          description: Success
+        default:
+          $ref: '#/components/responses/error'
+      x-ms-docs-operation-type: operation
+  '/groups/{group-id}/members/{directory-object-id}/$ref':
+    delete:
+      tags:
+        - group
+      summary: Delete member from a group
+      operationId: DeleteMember
+      parameters:
+        - name: group-id
+          in: path
+          description: 'key: id of group'
+          required: true
+          schema:
+            type: string
+          x-ms-docs-key-type: group
+        - name: directory-object-id
+          in: path
+          description: 'key: id of group member to remove'
+          required: true
+          schema:
+            type: string
+        - name: If-Match
+          in: header
+          description: ETag
+          schema:
+            type: string
+      responses:
+        '204':
+          description: Success
+        default:
+          $ref: '#/components/responses/error'
+      x-ms-docs-operation-type: operation
   /me/drive:
     get:
       tags:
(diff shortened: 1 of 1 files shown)


micbar:
shouldn't we be able to create a key `"members@odata.bind"` which has a collection of directory objects as a value?

micbar (approved):

rhafer:
> shouldn't we be able to create a key `"members@odata.bind"`

@micbar hm, true. That should work. For some reason I thought the body for the PATCH on `/groups/{id}` required some additional attributes.Please review once more.

I still think the API for managing group memberships as defined in ms graph is not exactly nice. I am trying to follow it anyway here.

> which has a collection of directory objects as a value?

Not really directory objects, but strings/uris referencing directory objects.

refs (approved):

api/openapi-spec/v0.0.yaml
```diff
@@ -498,6 +498,68 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
+  '/groups/{group-id}/members/$ref':
+    post:
+      tags:
+        - group
+      summary: Add a member to a group
+      operationId: AddGroupMember
```
refs, line 506: just a nitpick: I think `AddMember` reads better and could potentially lead to more concise code once this is generated into an sdk:

`group.AddGroupMember` vs `group.AddMember` kind of issue

api/openapi-spec/v0.0.yaml
```diff
@@ -498,6 +498,68 @@ paths:
...
+          description: Success
+        default:
+          $ref: '#/components/responses/error'
+      x-ms-docs-operation-type: operation
+  '/groups/{group-id}/members/{directory-object-id}/$ref':
+    delete:
+      tags:
+        - group
+      summary: Delete member from a group
+      operationId: DeleteGroupMember
```
refs, line 537: same as the above comment

refs (approved):

refs:
> shouldn't we be able to create a key `"members@odata.bind"` which has a collection of directory objects as a value?

@rhafer something like: https://swagger.io/docs/specification/data-models/data-types/#array

Also, mind this restriction by the msgraph spec:

> Note that up to 20 members can be added in a single request

The key would, obviously, be located in the `PATCH` request body

api/openapi-spec/v0.0.yaml
rhafer: Makes sense. I'll update that and then merge.

rhafer:
> > shouldn't we be able to create a key `"members@odata.bind"` which has a collection of directory objects as a value?
> 
> @rhafer something like: https://swagger.io/docs/specification/data-models/data-types/#array

Yeah that's what I did. :-)

> Also, mind this restriction by the msgraph spec:
> 
> > Note that up to 20 members can be added in a single request
> 
> The key would, obviously, be located in the `PATCH` request body

Yeah. Though that limit is not something I can express in this SPEC, is it? We might think about adding such a limit to the actual graph API implementation at some point though.

refs:
yes it can be done within the spec: https://swagger.io/docs/specification/2-0/describing-parameters/

> Additionally, you can:
> 
>     use minItems and maxItems to control the size of the array,
>     enforce uniqueItems,

```yaml
- in: query
          name: color
          required: false
          type: array
          minItems: 1
          maxItems: 5
          uniqueItems: true
          items:
            type: string
            enum: [black, white, gray, red, pink, orange, yellow, green, blue, purple, brown]
```

refs:
@rhafer  oh sorry, those docs are for v2. For v3: https://swagger.io/docs/specification/data-models/data-types/

image

rhafer:
Ok, thanks. I've added those limits to the spec. It doesn't seem to have any effect on the generated go code for the client. But the limit itself needs to be implemented server side anyway.

PR #32, opened 2022-02-24
ad more quota attributes

micbar:
# Description

Add more quota attributes.

`restricted` -> the fixed value for a quota limit.

@C0rby @butonic @felix-schwarz @JammingBen @TheOneRing

### DIFF
--- a/api/openapi-spec/v0.0.yaml
+++ b/api/openapi-spec/v0.0.yaml
@@ -1444,8 +1444,7 @@ components:
           description: The unique identifier for this item in the /drive/special collection
     quota:
       type: object
-      description: Optional. Information about the drive's storage space quota. Read-only.
-      readOnly: true
+      description: Optional. Information about the drive's storage space quota.
       properties:
         deleted:
           type: integer
@@ -1461,9 +1460,16 @@ components:
           type: string
           description: Enumeration value that indicates the state of the storage space. Read-only.
           readOnly: true
+        isRestricted:
+          type: boolean
+          description: Indicates if the maximum avatilable quota has been set.
+        restricted:
+          type: integer
+          description: Total allowed storage space, in bytes.
+          format: int64
         total:
           type: integer
-          description: 'Total allowed storage space, in bytes. Read-only.'
+          description: 'Dynamic value. Shows either the total possible available storage space or quota limit in bytes. Read-only.'
           format: int64
           readOnly: true
         used:


api/openapi-spec/v0.0.yaml
```diff
@@ -1461,9 +1461,16 @@ components:
           type: string
           description: Enumeration value that indicates the state of the storage space. Read-only.
           readOnly: true
+        isRestricted:
+          type: boolean
+          description: Indicates if the maximum avatilable quota has been set.
```
micbar, line 1466:
```suggestion
          description: Indicates if the maximum available quota has been set.
```

JanAckermann (approved):
🤝

api/openapi-spec/v0.0.yaml
```diff
@@ -1461,9 +1460,16 @@ components:
           type: string
           description: Enumeration value that indicates the state of the storage space. Read-only.
           readOnly: true
+        isRestricted:
```
butonic, line 1463: The correct way to delete the quota of a space / drive is to send a DELETE request to the property URL, eg to unset the total of a drive you would send `DELETE v1.0/drives/c12644a14b0a7750/quota/total`. I prefer using existing well defined ODATA semantics insdead of inventing unneeded properties.

micbar: ok, we can do that.
But we really need to know if the "total" is due to some restriction or due to lack of more storage.

api/openapi-spec/v0.0.yaml
```diff
@@ -1461,9 +1460,16 @@ components:
...
+          type: boolean
+          description: Indicates if the maximum available quota has been set.
+        restricted:
```
butonic, line 1466: is this supposed to give end users a way to see how many bytes are available on a storage provider (which is responsible to handle multiple spaces)?

I would not try to use the spaces / drives api to manage storage spaces. If we want to build a UI that allows space managers to creat new spaces, they need to see a list of storage providers, the number of spaces or the number of free bytes per storage space.

So, I vote against mix these two different levels of abstraction. IMO storages are a different resource than spaces.

api/openapi-spec/v0.0.yaml
JanAckermann: Added a ticket that describes the actual problem https://github.com/owncloud/ocis/issues/3233
@butonic

api/openapi-spec/v0.0.yaml
micbar: yes, therefore we need to separate it.

api/openapi-spec/v0.0.yaml
JanAckermann: Simple problem is, I set a quota to 200GB, but the available disk size is only 100GB. 
As a user I will loose the information which quota I had actually set.

butonic: `total` quota should not be reduced by actually available bytes. It is a static configuration property.

If the disk is running full the `remaining` property of the quota property will be less than `total`-`used`

JanAckermann: Well then it is not correctly implemented

butonic: https://github.com/owncloud/ocis/blob/77665a85d50f19cf9ea13140426a1f1e77409eb1/graph/pkg/service/v0/drives.go#L567-L576

correct, but we can put the missing data points into the opaque of https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.GetQuotaResponse or update the CS3 api right away.

micbar:
let me change my proposal

api/openapi-spec/v0.0.yaml
micbar: When you do not set a quota, the `total` value is basically unrestricted, which does not mean unlimited. On the graph API we need to know if a total quota is the value from the setting of the quota attribute or if it is smaller due to the hard limit of the storage. For example you set 5GB but on the CS3 api you only get 3GB because o storage limitiations.

If you set quota in the Web Client you have a dropdown list with values and the currently selected (which should be the real value on the backend) needs to be preselected. Currently the web Client cannot rely on that information because there is no distinction between the restriction from the limit or from the storage. 

It also makes no sense to show more "total" than really available. Showing 3GB total and not being able to upload 3GB would surely be considered as a bug.

micbar: Bottom line is: `total` always needs to be the real limit and the Clients also need to know the stored restriction value from the backend. So we need to split.

api/openapi-spec/v0.0.yaml
micbar: I disagree with the "remaining" That would be totally confusing for users if "used" + "remaining" != "total". So "total" needs to show the actually available if less than the setting.

api/openapi-spec/v0.0.yaml
butonic: I disagree. `quota.total` is the configured quota, or rather the *total* size of the space. if a client needs to determine if it can upload 3 GB it needs to look at `quota.remaining`.

butonic: from https://docs.microsoft.com/en-us/graph/api/resources/quota?view=graph-rest-1.0#properties
<html>
<body>


Property name | Type | Description
-- | -- | --
total | Int64 | Total allowed storage space, in bytes. Read-only.
used | Int64 | Total space used, in bytes. Read-only.
remaining | Int64 | Total space remaining before reaching the quota limit, in bytes. Read-only.
deleted | Int64 | Total space consumed by files in the recycle bin, in bytes. Read-only.


</body>
</html>

butonic: IMO quota has nothing to do with the actual free bytes. In oc10 this led to using negative values to represent certain corner cases:
```php
	/**
	 * @const \OCP\Files\FileInfo::SPACE_NOT_COMPUTED Return value for a not computed space value
	 * @since 8.0.0
	 */
	public const SPACE_NOT_COMPUTED = -1;
	/**
	 * @const \OCP\Files\FileInfo::SPACE_UNKNOWN Return value for unknown space value
	 * @since 8.0.0
	 */
	public const SPACE_UNKNOWN = -2;
	/**
	 * @const \OCP\Files\FileInfo::SPACE_UNKNOWN Return value for unlimited space
	 * @since 8.0.0
	 */
	public const SPACE_UNLIMITED = -3;
```

There are several places where the free space is calculated by substracting the used bytes from the quota value ... which is a lie in the first place if the remaining number of free bytes is different. Which can happen for multiple reasons.

The correct way to handle this is to not provide calculated values
- `total`: either there is a quota set, then `total` reflects that quota, or there is not, then `total` should be omitted. 
- `remaining` is either the available free bytes or total, whichever is lower
- `used` is the tree size
- `deleted` is the size of the trash

If `remaining` is calculated as `total-used` it would be redundant and should be dropped.

When a space has 'unlimited' quota in oc10 we return the available free disk space, which literally is the meaning of `remaining`.

Again, I think the problem is with the CS3 api which currently only knows `total` and `used`. It should support `remaining` as a third property.

butonic: what exactly can be expressed with `restricted` that cannot be expressed with these three
- `total` = the number of bytes you are entitled to (the configured quota or omitted if not set)
- `used` = the number of bytes you are using
- `remaining` = the number of bytes that are available (either free disk space or `total-used`, whichever is less, or if no total quota is set ... the free disk space. and we could omit the remaining bytes to represent unlimited)

`total` should not change when the disk has been filled up by external forces (log overflow, whatever), because the number of bytes the user is entitled to never changed. the admin has to ensure there is enough free space for the users.

micbar:
ok, i agree. We need to fix the `total` quota in reva.

pmaier1:
> IMO quota has nothing to do with the actual free bytes.

I agree. Please keep those separate. If the physical storage is full, you cannot upload more and will see errors but this is independent of the quota a user has. Storage needs separate monitoring.

Issue #30, opened 2022-02-17
Get rid of our fork of openapi-generator


dschmidt:
Currently we have to use a fork of openapi-generator (hosted at https://github.com/owncloud-ci/openapi-generator).
Check information in README if it's still needed or if we can archive the repo.

PR #33, opened 2022-03-02
Switch back to upstream openapi-generator

dschmidt:
Unfortunately we still need to use latest, as the fix from our fork is merged upstream but not released yet.

TODO in follow up PR: Pin version.

FIXES: https://github.com/owncloud/libre-graph-api/issues/30

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -27,9 +27,7 @@ config = {
 			'branch': 'main',
 		},
 	},
-	# FIXME: switch back to openapitools/openapi-generator-cli
-	# when https://github.com/OpenAPITools/openapi-generator/pull/11490 is merged
-	'openapi-generator-image': 'owncloudci/openapi-generator'
+	'openapi-generator-image': 'openapitools/openapi-generator-cli:latest'
 }
 
 def main(ctx):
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -22,14 +22,12 @@ config = {
...
-			'clean-src': "out-cpp-qt-client/client/*",
+			'clean-src': "out-cpp-qt-client/*",
 			'repo-slug': "libre-graph-api-cpp-qt-client",
 			'branch': 'main',
 		},
 	},
-	# FIXME: switch back to openapitools/openapi-generator-cli
-	# when https://github.com/OpenAPITools/openapi-generator/pull/11490 is merged
-	'openapi-generator-image': 'owncloudci/openapi-generator'
+	'openapi-generator-image': 'openapitools/openapi-generator-cli:latest'
```
micbar, line 30: could you pin that to the current commID?

micbar (changes requested):
Let us pin the version

.drone.star
dschmidt: No. I don't think they publish docker images for every commit:

https://hub.docker.com/r/openapitools/openapi-generator-cli/tags

micbar: ok then let us pin 5.4.0

micbar (approved):

PR #35, opened 2022-03-04
Qt: Allow to build without exceptions enabled

TheOneRing:


dschmidt (commented):
Hum ... exceptions are catchable, this isn't ... can we return an error instead?

dschmidt:
yeah ... well ...

dschmidt (approved):

PR #36, opened 2022-03-07
add driveAlias

micbar:
# Description

According to https://owncloud.dev/ocis/adr/0011-global-url-format/#mixed-global-urls we can use a human-readable space alias to route space in clients.

The alias is a string like `personal/einstein` or `projects/mars`.
This alias is not enforced to be unique. The main identifier will still be the unique ID.

URLs will always contain both 1) alias and 2) ID


kulmann (approved):
👍

PR #43, opened 2022-05-12
Bump version to v1.14.2

rhafer:
LICENSE was still being	removed in target repos. Should be fixed with
latest change in .drone.star


rhafer:
@micbar We need yet another version bump. Let's hope this is working now and the license files stay in place in the target repos  :pray:

micbar (approved):

PR #61, opened 2022-12-13
clean up unused props and add examples for drives

micbar:
# Description

## Clean up

- unused parameters
- better tags

## Required

- Add required body fields to the drive entity


micbar:
@kulmann I hope the unused parameters do not break the upgrade in web. I see that web is sending some of them by default.

butonic (approved):

Issue #59, opened 2022-12-09
[Enhancement] Add libre tags API


micbar:
# Description

We need to add the tags assign / unassign / list endpoints.

## Proposal
Add a new `tags` resource in a non-colliding way:
`https://ocis.test/graph/v1.0/extensions/org.libregraph/tags/`

Note: `https://ocis.test/graph/v1.0/extensions` is not known to MS Graph, but we apply the [open extension mechanism](https://learn.microsoft.com/en-us/graph/extensibility-overview?context=graph%2Fapi%2F1.0&preserve-view=true&tabs=http#open-extensions) to the root resource to add a custom `tags` resources to the graph.

### Tag Scheme

| name | type |
|------|-----|
| name | string* |

* lowercase

### Operations

- List tags
- Assign tags
- Unassign tags

PR #62, opened 2022-12-14
add tags endpoints

fschade:
add graph endpoints for tags
fixes https://github.com/owncloud/libre-graph-api/issues/59

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -1595,14 +1595,7 @@ paths:
         content:
           application/json:
             schema:
-              type: object
-              properties:
-                resourceId:
-                  type: string
-                tags:
-                  type: array
-                  items:
-                    type: string
+              $ref: '#/components/schemas/tagAssignment' 
     delete:
       tags:
         - tags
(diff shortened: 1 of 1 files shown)


micbar (approved):
Please add two examples for the put and delete requests

micbar (changes requested):
There is a way to mark body properties required.

fschade:
@micbar can you re-check if the examples are correct?

micbar:
https://github.com/owncloud/libre-graph-api/blob/main/api/openapi-spec/v1.0.yaml#L1606-L1608

api/openapi-spec/v1.0.yaml
```diff
@@ -1565,6 +1565,57 @@ paths:
           description: Success
         default:
           $ref: '#/components/responses/error'
+  '/extensions/org.libregraph/tags':
```
kulmann, line 1568: we can't register the domain `libregraph.org` (already taken). I'd recommend registering e.g. `libregraph.io` and use the namespace accordingly.

fschade: done

fschade: reverted, back to org.libregraph

micbar (approved):

Issue #38, opened 2022-04-14
Pin openapi-generator to release version


dschmidt:
openapi-generator [v6.0.0-beta](https://github.com/OpenAPITools/openapi-generator/releases/tag/v6.0.0-beta) was released with my CPP fix in it.

We should update and pin to 0.6.0 final when it's released.

PR #78, opened 2023-01-04
Bump openapi-generator to v6.2.0

rhafer:
We now pin the image to the SHA for the 6.2.0 release. To get away from the pre-release snapshot we've been using for a while now.

Closes: #38

As far as I could see this doesn't introduce any problematic changes to the libraries that we are generating from this.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -24,7 +24,7 @@ config = {
 			'branch': 'main',
 		},
 	},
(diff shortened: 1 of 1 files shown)


rhafer:
Note: I've downgraded this to v6.2.0 since the go test generation introduced in v6.2.1 has an issue that makes golangci-lint fail. The fix is in: https://github.com/OpenAPITools/openapi-generator/pull/13905 (will be part of 6.3.0)

dschmidt (approved):

Issue #25, opened 2022-01-31
Difference between drive.createdBy and drive.createdByUser?


felix-schwarz:
The schema for `drive` describes `createdBy` and `createdByUser` as follows:
- `createdBy`: `identitySet`, `Identity of the user, device, or application which created the item. Read-only.`
- `createdByUser` : `user`, Identity of the user who created the item. Read-only.`

The information here seems to be redundant, as the meaning is described as the same and `identitySet` can also store a `user`. Why does `drive.createdByUser` exist and when should it be used?

butonic:
`createdByUser` does not exist on the graph api, it should be removed from the libregraph.

rhafer:
> `createdByUser` does not exist on the graph api, it should be removed from libregraph.

Agreed. And since we're neither using ` createdBy` nor `createdByUser` anywhere in the  server yet. I think it's save to just remove `createdByUser` from the spec. The same is btw true for `lastModifiedByUser` vs `lastModifiedBy`.

PR #81, opened 2023-01-05
Remove redundant and unused 'createdByUser' and 'lastModifiedByUser'

rhafer:
Not sure how that ended up in our spec but we're not using them currently and the graph api mandates createdBy and lastModifiedBy instead.

Closes: #25


butonic:
AFAICT the ms graph item has `createdByUser` AND `createdBy` on `driveItem`, but not on `drive`. which does not answer the original question. I'd go with `createdBy` only on both.

PR #89, opened 2023-01-17
add an expiration date time to the permission schema

C0rby:
Required to list the space member expirations.


api/openapi-spec/v1.0.yaml
```diff
@@ -2584,6 +2584,9 @@ components:
       description: 'The Permission resource provides information about a sharing permission granted for a DriveItem resource.'
       readOnly: true
       properties:
+        expirationDateTime:
+          type: string
```
micbar, line 2588: `string` , yes. If you compare it e.g. with  other date fields you can add a pattern and format to specify it further.

rhafer (approved):

Issue #72, opened 2022-12-20
Flesh out education classes endpoints


rhafer:
We're at least lacking:

GET/POST/DELETE  (and PATCH?)  on /education/classes/{id}/teachers

PR #90, opened 2023-01-19
Add GET/POST/DELETE for managing teachers for classes.

Excds:
This adds the missing endpoints for managing teachers for classes.

Fixes https://github.com/owncloud/libre-graph-api/issues/72

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -1754,6 +1754,108 @@ paths:
           description: Success
         default:
           $ref: '#/components/responses/error'
+  '/education/classes/{class-id}/teachers':
+    get:
+      tags:
+        - educationClass.teachers
+      summary: Get the teachers for a class
+      security:
+        - bearerAuth: []
+      operationId: GetTeachers
+      parameters:
+        - name: class-id
+          in: path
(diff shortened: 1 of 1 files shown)


api/openapi-spec/v1.0.yaml
```diff
@@ -1754,6 +1754,108 @@ paths:
...
+          required: true
+          schema:
+            type: string
+            example: 86948e45-96a6-43df-b83d-46e92afd30de
+      requestBody:
+        description: educationUser to be added as teacher
+        content:
+          application/json:
+            schema:
+              title: Class Teacher Reference
```
butonic, line 1820: @rhafer do we want to allow username aka issuer assigned identifier?

rhafer: @butonic Username (i.e. `onPremisesSamAccountName`) yes. As we support that for  `/education/classes/{class-id}/members/$ref`  and `/education/group/{id}/members/$ref` already. Regarding the `identities` (I guess that is what you're referring to with `issuer assigned identifier`), I am still undecided. But does that even need to be expressed in the spec? If yes, how?

rhafer (approved):

PR #91, opened 2023-01-25
Improve the CMakeLists Template

dragotin:
Add the ability to build both static and shared libraries which will ease upstream packaging.

Also a bit more useful README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,15 @@
 # Libre Graph Api
 
-An API for open Cloud Collaboration.
+An API for open Cloud Collaboration. See the [Libre Graph Home](https://libregraph.github.io/) for more details.
 
 This API is inspired by [Microsoft Graph API](https://developer.microsoft.com/en-us/graph).
 
+
 ## Goal
 
-This projects goal is to provide an open source standard for open Cloud Collaboration.
+The project goal is to provide an open source standard for open Cloud Collaboration.
+
+Libre Graph is open source and open to any open source project that implements endpoints of the API.
 
 ## Specification
 
@@ -22,7 +25,13 @@ You can find a rendered version of the [API documentation](https://owncloud.dev/
 
 ## Clients
 
-Client code can be generated from the API spec. 
+Client code can be generated from the API spec.
+
+For example, to run the generator for the C++ bindings locally, run the following docker based command:
+```bash
+docker run --rm -v "${PWD}:/local" openapitools/openapi-generator-cli generate --enable-post-process-file  -t local/templates/cpp-qt-client  -i local/api/openapi-spec/v1.0.yaml -g cpp-qt-client -o /local/out/cpp
+```
+That generates the output in out/cpp.
 
 
 ### Available client libraries
--- a/templates/cpp-qt-client/CMakeLists.txt.mustache
+++ b/templates/cpp-qt-client/CMakeLists.txt.mustache
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.4)
 string(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]" APP_VERSION "{{appVersion}}")
 string(REGEX REPLACE " " "" TARGET_NAME "{{appName}}")
 
@@ -7,6 +7,8 @@ project("{{appName}}" VERSION ${APP_VERSION} LANGUAGES CXX)
 set(CMAKE_VERBOSE_MAKEFILE ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_AUTOMOC ON)
+
+include(GNUInstallDirs)
 
 if (MSVC)
     string(REGEX REPLACE "/W[1-3]" "/W4" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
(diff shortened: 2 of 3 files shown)


templates/cpp-qt-client/CMakeLists.txt.mustache
```diff
@@ -1,6 +1,7 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.4)
 string(REGEX MATCH "[0-9]+\\.[0-9]+\\.[0-9]" APP_VERSION "{{appVersion}}")
-string(REGEX REPLACE " " "" TARGET_NAME "{{appName}}")
+
+set(TARGET_NAME "LibreGraphAPI")
```
micbar, line 4: we could keep this generic as it was. This template would stay usable for other repos if we keep the regex replace

micbar (approved):
Only a small nitpick 😄

README.md
```diff
@@ -22,7 +25,13 @@ You can find a rendered version of the [API documentation](https://owncloud.dev/
 
 ## Clients
 
-Client code can be generated from the API spec. 
+Client code can be generated from the API spec.
+
+For example, to run the generator for the C++ bindings locally, run the following docker based command:
+```bash
+docker run --rm -v "${PWD}:/local" openapitools/openapi-generator-cli generate --enable-post-process-file  -t local/templates/cpp-qt-client  -i local/api/openapi-spec/v1.0.yaml -g cpp-qt-client -o /local/out/cpp
```
dschmidt, line 32: 👍️

templates/cpp-qt-client/CMakeLists.txt.mustache
dschmidt: 👍️

templates/cpp-qt-client/CMakeLists.txt.mustache
```diff
@@ -75,5 +80,5 @@ set(HEADER
     {{prefix}}Object.h
 )
 install(FILES ${HEADER} DESTINATION include/OpenAPI/${TARGET_NAME})
-install(TARGETS client EXPORT ${TARGET_NAME}Config RUNTIME DESTINATION bin LIBRARY DESTINATION lib ARCHIVE DESTINATION lib)
-install(EXPORT ${TARGET_NAME}Config DESTINATION lib/cmake/${TARGET_NAME} NAMESPACE OpenAPI::)
+install(TARGETS client EXPORT ${TARGET_NAME}Config RUNTIME DESTINATION bin LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+install(EXPORT ${TARGET_NAME}Config DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${TARGET_NAME} NAMESPACE OpenAPI::)
```
dschmidt, line 84: I think you need include GNUInstallDirs to make these vars available outside of packaging contexts

templates/cpp-qt-client/CMakeLists.txt.mustache
```diff
@@ -46,6 +47,10 @@ add_library(OpenAPI::${TARGET_NAME} ALIAS client)
 
 target_link_libraries(client PRIVATE Qt5::Core Qt5::Network{{#authMethods}}{{#isOAuth}} Qt5::Gui{{/isOAuth}}{{/authMethods}}{{#contentCompression}} ${ZLIB_LIBRARIES}{{/contentCompression}})
 
+if (NOT WIN32)
```
dschmidt, line 50: What is the reason for this platform guard?

templates/cpp-qt-client/CMakeLists.txt.mustache
```diff
@@ -19,7 +20,7 @@ find_package(Qt5Network REQUIRED){{#authMethods}}{{#isOAuth}}
 find_package(Qt5Gui REQUIRED){{/isOAuth}}{{/authMethods}}{{#contentCompression}}
 find_package(ZLIB REQUIRED){{/contentCompression}}
 
-add_library(client STATIC
+add_library(client
```
TheOneRing, line 23: Can we keep it static for Windows?
The project has no exports and can only be build static on Windows.

dragotin: Sure, just use `BUILD_SHARED_LIBS` in calling cmake, as described here: https://stackoverflow.com/questions/61386679/how-to-get-static-library-default-by-cmake

dragotin: Please elaborate. The current state builds a static library nicely out of the box on Linux. How is that different for Windows?

TheOneRing: On Windows symbols need to be exported using a macro.
On linux this would be symbols visibility. Linux defaults to everything is visible.
As those macros are missing in the generated code, it wont link or at least the resulting library would be unlinkable.

dragotin: I know. 

The whole point of this PR is to be able to build both, static and shared libraries as prefered on Linux and needed for packaging. If the STATIC is kept with in `add_library`, I did not succeed to overwrite it to create a shared lib. Maybe you know how to do it?

Omitting the STATIC allows to simply choose which type of lib should be built.

Also, please see:
grafik
from https://matgomes.com/add-library-cmake-create-libraries/

TheOneRing: Maybe we could just do 
```
# On  Windows only static builds are supported.
set(BUILD_SHARED_LIBS OFF)
```

templates/cpp-qt-client/CMakeLists.txt.mustache
dragotin: Added that

templates/cpp-qt-client/CMakeLists.txt.mustache
dragotin: Added. Pls approve.

TheOneRing (approved):

PR #104, opened 2023-04-26
Qt6 support

TheOneRing:
Depends on https://github.com/owncloud/libre-graph-api/pull/103

@dragotin this will prefer Qt6, from your perspective as a packager, should I add an option to force Qt6 and assume 5 otherwise?


fmoc (approved):

dragotin:
I think the way it is now is ok. Thanks!

PR #106, opened 2023-04-27
Clean generated directory before we update it

TheOneRing:
Without this step orphaned files will stay forever

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -157,7 +157,9 @@ def generate(ctx, lang):
 				'pull': 'always',
 				'commands': [
 					'test -d "templates/{0}" && TEMPLATE_ARG="-t templates/{0}" || TEMPLATE_ARG=""'.format(lang),
+					'rm -Rf %s/*' % config["languages"][lang]["src"],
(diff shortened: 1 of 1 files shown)


TheOneRing:
Its the approach we use for the desktop translations and it got us rid of 10 year old outdated translations.
However in this case we have a mix of generated code and manually added code.
The repos dont have a common generation dir we could purge so I'm not sure.

For now I could purge the client repo of the c++ bindings, we should really get rid of the binary artifacts from the validation step...
https://github.com/owncloud/libre-graph-api-cpp-qt-client/blob/main/client/libLibreGraphAPI.a

fmoc:
Can't tell whether this approach makes any sense, really. I was not involved in setting this up.

Can the "do not commit generated files" part be fixed by defining ignore patterns very carefully (e.g., by using an allowlist approach rather than the default excludelist one)? Or can we just fix the generator to add support for out-of-source builds?

TheOneRing:
But orphaned files would still remain.

TheOneRing:
We now purge the generated code and copy the licence from the main repo to the generated code.

micbar (approved):

PR #109, opened 2023-05-09
add space root to remote item

micbar:


api/openapi-spec/v1.0.yaml
```diff
@@ -2551,6 +2551,15 @@ components:
           $ref: '#/components/schemas/fileSystemInfo'
         folder:
            $ref: '#/components/schemas/folder'
+        driveAlias:
+          type: string
+          description: "The drive alias can be used in clients to make the urls user friendly. Example: 'personal/einstein'. This will be used to resolve to the correct driveID."
+        path:
+          type: string
+          description: "The relative path of the item related to it's drive root."
```
kulmann, line 2559:
```suggestion
          description: "The relative path of the item in relation to its drive root."
```

butonic:
Can you at least link to the issue you are trying to solve? I wonder if this change is really necessary. If you leave out the node it of a remote item it will give you the space id...

PR #110, opened 2023-06-27
Bump openapi-generator to v6.6.0

rhafer:
The image is now pinned to the sha256 of v6.6.0.

This gets use the latest  fixes for the golang generated code.


rhafer:
Setting this to draft for now. There seem to be some issues introduced in the golang generator. (At least the unit test on the graph service fail with the new client code)

PR #112, opened 2023-08-09
[WIP] add Sharing NG endpoints

butonic:
libregraph API spec for the sharing ADR in https://github.com/owncloud/ocis/pull/6995

When following a share lifecycle, the following OCS requests can be replaced with these libregraph counterparts:

## List shares on a resource (includes sharing options)
**OCS**: web calls two endpoints:
- `/ocs/v1.php/apps/files_sharing/api/v1/shares?path={space-relative-path}&space_ref={space-id}&reshares=true`
- `/ocs/v1.php/apps/files_sharing/api/v1/shares?path={space-relative-path}&space_ref={space-id}&shared_with_me=true`

**ms graph**: `/drives/{drive-id}/items/{drive-item-id}/permissions`
```json
{
    "value": [
        {
            "id": "collaborative-share-id",
            "roles": [
                "write"
            ],
            "grantedToV2": {
                "user": {
                    "displayName": "Albert Einstein",
                    "id": "4c510ada-c86b-4815-8820-42cdf82c3d51"
                }
            },
        },
        {
            "id": "public-link-share-id",
            "roles": [
                "read"
            ],
            "link": {
                "webUrl": "https://cloud.example.com/s/CMXOrzoFODpHKsS"
            }
        }
    ]
}
```
**libre graph**: for now we stick to two kinds of permissions:
* `grantedToV2` for internal user or group shares
* `link` for public link shares
_jfd: roles are currently hardcoded to `read`, `write` and `owner`. We can make them customizable but I'll show that at the end of this description_
```json
{
    "value": [
        {
            "id": "collaborative-share-id",
            "roles": [
                "write"
            ],
            "grantedToV2": {
                "user": {
                    "displayName": "Albert Einstein",
                    "id": "4c510ada-c86b-4815-8820-42cdf82c3d51"
                }
            },
        },
        {
            "id": "public-link-share-id",
            "roles": [
                "read"
            ],
            "link": {
                "webUrl": "https://cloud.example.com/s/CMXOrzoFODpHKsS"
            }
        }
    ]
}
```

## Search for recipient (not part of this PR)
**OCS**: `/ocs/v2.php/apps/files_sharing/api/v1/sharees?search=einst&itemType=(folder|file)&page=1&perPage=200&format=json`
**ms graph**: [`/me/people?$search="einst"`](https://learn.microsoft.com/en-us/graph/api/user-list-people?view=graph-rest-beta&tabs=http) is used to interact with users that are [relevant or in a working-with relationship](https://learn.microsoft.com/en-us/graph/people-insights-overview#including-a-person-as-relevant-or-working-with). The returned list of [person](https://learn.microsoft.com/en-us/graph/api/resources/person?view=graph-rest-beta) entities has a  [`personType` property](https://learn.microsoft.com/en-us/graph/search-concept-person#person-types) to differentiate types of groups and users.
**libre graph**: `/me/people?$search="einst"` is the only endpoint we need, I think. For now, a `$search="einst"` parameter can be used instead of the OCS sharee call.

**Note**: There is a difference between `$search="foo bar"` and `$search=foo bar`. To mimic OCS the request has to quote the typed in string.*

## Create a share with user or group
**OCS**:  
```
POST ocs/v1.php/apps/files_sharing/api/v1/shares

shareType=0
shareWith=marie
path=/Neuer Ordner
space_ref=storage-users-1$some-admin-user-id-0000-000000000000!71beebf5-0057-4104-a814-bb49712eaab9
permissions=31
role=editor
```

**ms graph**: sharing wit a user is done by posting an `invite`
```json
POST /drives/{drive-id}/items/{drive-item-id}/invite

{
    "requireSignIn": true,
    "recipients": [
        {
            "email": "einstein@example.org"
        }
    ],
    "roles": [
        "read"
    ]
}
```

response:
```json
{
    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(permission)",
    "value": [
        {
            "@odata.type": "#microsoft.graph.permission",
            "id": "r_mJa3kBqBtkotIl8tWt9nVA2L1",
            "roles": [
                "read"
            ],
            "shareId": "s!BFB3CkuhRCbBgmcZvWieFTBrBGQ",
            "expirationDateTime": "0001-01-01T00:00:00Z",
            "hasPassword": false,
            "grantedToV2": {
                "user": {
                    "id": "einstein@example.org"
                }
            },
            "invitation": {
                "email": "einstein@example.org",
                "signInRequired": true
            },
            "link": {
                "webUrl": "https://1drv.ms/f/s!BFB3CkuhRCbBgmcZvWieFTBrBGQ"
            }
        }
    ]
}
```
It seems the response does not use `grantedToV2`. And when `requireSignIn` is left out or `false` it will happily create a link `webUrl` that allows browsing the file. 

**libre graph**: sharing wit a user is done by posting an `invite`. While ms graph uses `email` to identify a recipient libre graph assumes internal shares are created using the `objectId` (the users `id`) 

```json
POST /drives/{drive-id}/items/{drive-item-id}/invite

{
    "requireSignIn": true,
    "recipients": [
        {
            "objectId": "4c510ada-c86b-4815-8820-42cdf82c3d51"
        }
    ],
    "roles": [
        "read"
    ]
}
```

response:
```json
{
    "value": [
        {
            "id": "8c5ed185-1fcb-4c5a-8569-b9ed04293204",
            "roles": [
                "read"
            ],
            "grantedToV2": {
                "user": {
                    "id": "4c510ada-c86b-4815-8820-42cdf82c3d51"
                }
            },
        }
    ]
}
```
_jfd: multiple recipients can be sent in the same request. Each will receive a dedicated permissions object with their own id. WChen an error occurs a 207 multistatus response will be returned, similar to https://learn.microsoft.com/en-us/graph/api/site-follow?view=graph-rest-1.0&tabs=http#response-1 which shows an examplo where one of the entries contains an error_
request:
```json
POST /drives/{drive-id}/items/{drive-item-id}/invite

{
    "recipients": [
        {
          "objectId": "4c510ada-c86b-4815-8820-42cdf82c3d51"
        },
        {
          "objectId": "f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c"
        }
    ],
    "roles": [
        "read"
    ]
}
```

response:
```json
{
  "value": [
    {
      "id": "81d5bad3-3eff-410a-a2ea-eda2d14d4474",
      "roles": [
        "write"
      ],
      "grantedToV2": [
        {
          "user": {
            "id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
            "displayName": "Albert Einstein"
          }
        }
      ]
    },
    {
      "id": "b470677e-a7f5-4304-8ef5-f5056a21fff1",
      "error": {
        "@odata.type": "#odata.error.main",
        "code": "invalidRequest",
        "message": "The user id that is provided in the request is incorrect",
        "innerError": {
            "code": "itemNotFound",
            "errorType": "expected",
            "message": "Unknown user id b470677e-a7f5-4304-8ef5-f5056a21fff1 ",
        }
      }
    }
  ]
}
```

## Shared by me
List shares created by the currently logged in user
**OCS**: is called Shared With Others, see below
**ms graph**: odata query not implemented
**onedrive web**: can list shares by me in the web ui. maybe something is coming to ms graph as well?
**libre graph**: `/me/drive/sharedByMe` (only includes shares created by the current user)

## Shared with others (not part of this PR)
Includes shares in project spaces that can be managed by the current user because he is a manager of the space. He does not need to be the creator of the share.
**OCS**: `/ocs/v1.php/apps/files_sharing/api/v1/shares?reshares=true&include_tags=false&share_types=0,1,4,6`
**ms graph**: odata query not implemented
**libre graph**: does not exist, but we could add a new `/me/drive/sharedWithOthers` endpoint

***Note**: the OCS `share_types` are 0 = User, 1 = Group, 3 = public link, 4 = guest, 6 = federated share, 7 = federated group / space member user (OH great we have a clash here), 8 = space member group.*

***Hint**: There is a difference in the concept of shares. In OC10 shares are tied to a user. In OCIS they are tied to a space. The Owner/All managers of a space can collaboratively manage all shares in a personal/project space.*

## Shared with me
As the recipient you want to get a list of all driveItems that have been shared with you.
**OCS**: `/ocs/v1.php/apps/files_sharing/api/v1/shares?include_tags=false&state=all&shared_with_me=true`
**ms graph**: `/me/drive/sharedWithMe`
**libre graph**: `/me/drive/sharedWithMe`
The endpoint returns a `collection(driveItem)` that contains both: mounted and unmounted driveItems. shared driveItems are wrapped in another driveItem representing the mountpoint as on the ms graph API. But there are two deviations from the ms graph API: 
1. Shared driveItems that have not been mounted are **not** wrapped by another drive item. The ms graph api always wraps the shared driveItem. We do not want to do that on the fly as the driveItem representing the mountpoint would not have a parent id as it is unmounted. We will only wrap the shared driveItem with a mount point if the share has been created.
2. libre graph for now does not use the `shared` property but expands the `permissions` relation as it contains all necessary information. This might change if we think the `shared` property is enough to show the necessary indicators. It would be faster than listing all details of the permission.
3. libre graph uses a `@libre.graph.hidden` annotation on driveItems to indicate 
```json
{
  "value": [
    {
      "id": "u-u-id-of-mountpoint",
      "name": "November-December Ad Proposals.pptx",
      "size": 100984,
      "parentReference": {
        "driveType": "personal",
        "driveId": "u-u-id-of-drive-containing-the-pointpoint",
        "id": "parent-u-u-id"
      },
      "remoteItem": {
        "id": "u-u-id-of-shared-driveItem",
        "name": "November-December Ad Proposals.pptx",
        "size": 100984,
        "parentReference": {
          "driveType": "personal",
          "driveId": "u-u-id-of-drive-containing-the-item"
          // no "id" because recipient is not allowed to see parent
        },
        "permissions": [
          {
            "@libre.graph.hidden": false,
            "id": "92f276ac-827a-40c1-9d9c-bb41a628ce71",
            "roles": [
              "write"
            ],
            "grantedToV2": {
              "user": {
                "displayName": "Jörn Dreyer",
                "id": "c12644a14b0a7750"
              }
            }
          }
        ]
      }
    },
    {
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "size": 17776,
      "parentReference": {
        "driveType": "personal",
        "driveId": "u-u-id-of-drive-containing-the-item"
        // no "id" because recipient is not allowed to see parent
      },
      "permissions": [
        {
          "@libre.graph.hidden": false,
          "id": "477731b4-56a6-4f58-9530-2e08bdf52df5",
          "roles": [
            "write"
          ],
          "grantedToV2": {
            "user": {
              "displayName": "Jörn Dreyer",
              "id": "c12644a14b0a7750"
            }
          }
        }
      ]
    },
    {
      "id": "a7d033f9-6093-43bb-91a7-bc82265e6a7f",
      "name": "Irrelevant Marketing Term Successes International.xlsx",
      "size": 176536587,
      "parentReference": {
        "driveType": "personal",
        "driveId": "u-u-id-of-drive-containing-the-item"
        // no "id" because recipient is not allowed to see parent
      },
      "permissions": [
        {
          "@libre.graph.hidden": true,
          "id": "5277be2e-db85-4d11-9a6c-28853142f279",
          "roles": [
            "write"
          ],
          "grantedToV2": {
            "user": {
              "displayName": "Jörn Dreyer",
              "id": "c12644a14b0a7750"
            }
          }
        }
      ]
    }
  ]
}
```
_jfd: we cannot put the `@libre.graph.hidden` property on the drive item directly, because it would mean the driveItem was hidden. So we put it on the permissions property. That hidden flag can be toggled._

## Accept/Reject a share / an invite
**OCS**:
- Accept `POST ocs/v2.php/apps/files_sharing/api/v1/shares/pending/{share-id}`
- Deny `DELETE ocs/v2.php/apps/files_sharing/api/v1/shares/pending/{share-id}`
**ms graph**: /sharedWithMe will list all shares, but
- I cannot see a property that shows the mountpoint ... the parentReference only contains
```json
"parentReference": {
    "driveId": "c12644a14b0a7750",
    "driveType": "personal"
},
```
**libre graph**: mount shares in the share jail by creating a driveItem using a `POST /drives/{sharejailid}/items` request:
```json
{
  "name": "Einsteins project share",
  "remoteItem": {
    "id": "{drive-item-id}"
  }
}
```
This request can check if the current user has access to the given drive item and will 'mount'/ accept it with the name "Einsteins project share".  In theory we could add a remoteItem anywhere in a drive like in OC10, but it is a product decision to collect them in the dedicated virtual share jail drive.

The mountpoint can be deleted/rejected by sending a `DELETE /drives/{drive-item-id}` where `{drive-item-id}` is the id of the drive item representing the mount point, not the remote item.

## Create a public link
**ms graph**
```json
POST /drives/{drive-id}/items/{drive-item-id}/createLink

{
    "type": "view",
    "scope": "anonymous"
}
```
response:
```json
{
    "id": "{permission-id}",
    "link": {
        "type": "view",
        "webUrl": "https://1drv.ms/f/s!AlB3CkuhRCbBgmde472f8-qxYdg",
        "application": {
            "id": "4c1ad100"
        }
    }
}
```

## Update shares / links
**ms graph**:
_jfd: same as libregraph but I tried changing the role ... we'll just not go there for now_
The role does affect the ui. It now shows upload elements even for non logged in users. The link type is still view. When trying to upload a file the web ui will try to log you in before making the request. The link type cannot be changed: in the UI there is a hint "This setting can't be changed. Create a new link if you need different permissions.". I guess this is to prevent changing permissions on existing links?

**libre graph**:
```json
PATCH /drives/{drive-id}/items/{drive-item-id}/permissions/{permission-id}

{
    "link": {
        "type": "edit"
    }
}
```

response
```json
{
    "id": "{permission-id}",
    "link": {
        "type": "edit",
        "webUrl": "https://1drv.ms/f/s!AlB3CkuhRCbBgmde472f8-qxYdg",
        "application": {
            "id": "4c1ad100"
        }
    }
}
```
For links we ignore the role and instead set the link type. ms graph has a lot predefined: https://learn.microsoft.com/en-us/graph/api/listitem-createlink?view=graph-rest-beta&tabs=http#link-types
<html><body>

Type value | Description
-- | --
view | Creates a read-only link to the item.
review | Creates a review link to the item. This option is only available for files in OneDrive for Business and SharePoint.
edit | Creates an read-write link to the item.
embed | Creates an embeddable link to the item.
blocksDownload | Creates a read-only link that blocks  download to the item. This option is only available for files in  OneDrive for Business and SharePoint.
createOnly | Creates an upload-only link to the item. This option is only available for folders in OneDrive for Business and SharePoint.
addressBar | Creates the default link that is shown in  the browser address bars for newly created files. Only available in  OneDrive for Business and SharePoint. The organization admin configures  whether this link type is supported, and what features are supported by  this link type.
adminDefault | Creates the default link to the item as  determined by the administrator of the organization. Only available in  OneDrive for Business and SharePoint. The policy is enforced for the  organization by the admin

</body>
</html>

## Delete shares / links
**OCS** `DELETE ocs/v1.php/apps/files_sharing/api/v1/shares/{share-id}`
**libre graph** `DELETE /drives/{drive-id}/items/{drive-item-id}/permissions/{permission-id}`

# older notes

- [x] `/drives/{drive-id}/items/{item-id}/invite` to create shares
- [x] `/drives/{drive-id}/items/{item-id}/createLink` to create links
- [x] `/me/drive/sharedWithMe` to list shares with the current user
- [x] `/me/drive/sharedByMe` to list shares by the current user. This only includes shares created by the current user, not all shares he can manage, eg. in project spaces.
- [ ] ~~extend `/drives` to allow `/drives?$expand=items&$filter=items/any(property:property/shared+ne+null)`, the listing of shares with others~~ we went with a libregraph only `/me/drive/sharedByMe` endpoint instead
  - Q: should this work for all spaces or only on a per space/drive basis? in project spaces, does a user want to see all shares in the space or only the ones he created? This request lists any driveItem that is shared aka that has a [`shared` facet](https://learn.microsoft.com/en-us/graph/api/resources/shared?view=graph-rest-1.0)
- [x] `/drives/{drive-id}/items/{item-id}/permissions` to list permissions
  - [x] get permission
  - [x] update permission
  - [x] delete permission
- [x] How do we make [the list of preconfigured roles](https://learn.microsoft.com/en-us/graph/api/resources/permission?view=graph-rest-1.0#roles-property-values) configurable?
  - ~~sharing roles as app roles / [appRoleAssignment](https://learn.microsoft.com/en-us/graph/api/resources/approleassignment?view=graph-rest-1.0) for a sharing app???~~
  - ~~a `sharing` relationship on the [`roleManagement` resource](https://learn.microsoft.com/en-us/graph/api/resources/rolemanagement?view=graph-rest-1.0#relationships) seems more fitting. appRoles are supposed to show up in access token scopes ... so ... nah~~
  - [x] clients want to be able discover the sharing roles that can be used for a specific driveItem
     - uses a collectiong annotation in the list permissions response at `/drives/{drive-id}/items/{item-id}/permissions`:
       ```
       {
          "@libre.graph.permissions.roles.allowedValues": [
            {
              "name": "read",
              "displayname": "Viewer"
            },
            {
              "name": "write",
              "displayname": "Editor"
            },
            {
              "name": "owner",
              "displayname": "Manager"
            }
          ],
          "value": [
             ...
          ]
       }
       ```
  - [x] /roleManagement/permissions/roleDefinitions
- [ ] ms graph has `/drives/{drive-id}/items/{item-id}/invite` and `/drives/{drive-id}/items/{item-id}/createLink`. They both add permissions to drive items and the `permission` object as a result can carry a `link` or a `grantedToV2` facet/property. This needs more investigation: permissioens have a `role` property but `link` objects themselves have a `type` ... and the existing [link types in ms graph](https://learn.microsoft.com/en-us/graph/api/driveitem-createlink?view=graph-rest-beta&tabs=http#link-types) pretty much cover all we need. That being said we want to have them customizeable ... so they should be discoverable, similar to the sharing roles at `/roleManagement/permissions/roleDefinitions`

## Further reading

It follows the MS Graph API, documented here:
* shares in ms graph are called permissions: https://learn.microsoft.com/en-us/graph/api/resources/permission?view=graph-rest-1.0
* invite is used to add permissions and optionally send a message https://learn.microsoft.com/en-us/graph/api/driveitem-invite?view=graph-rest-1.0&tabs=http
* list permissions on a driveItem: https://learn.microsoft.com/en-us/graph/api/driveitem-list-permissions?view=graph-rest-1.0&tabs=http
* get permission: https://learn.microsoft.com/en-us/graph/api/permission-get?view=graph-rest-1.0&tabs=http
* update permission: https://learn.microsoft.com/en-us/graph/api/permission-update?view=graph-rest-1.0&tabs=http
* delete permission: https://learn.microsoft.com/en-us/graph/api/permission-delete?view=graph-rest-1.0&tabs=http

* sharing roles are only read, view, owner: https://learn.microsoft.com/en-us/graph/api/resources/permission?view=graph-rest-1.0#roles-property-values

Listing shares incoming / outgoing shares:

* list shared with me https://learn.microsoft.com/en-us/graph/api/drive-sharedwithme
* list shared with others has no official andpoint, but the semantically matching request might be `https://graph.microsoft.com/v1.0/me/drive/items?$filter=shared ne null`. It works by matching all driveItems that have no sharing facet. However, it is not implemented in ms graph. Maybe because it is not clear that `/v1.0/me/drive/items` works on ALL drives. Semantically, it should only list the shared drive items on the users personal drive. To match drive items on all drives the user has access to `https://graph.microsoft.com/v1.0/drives?$expand=items&$filter=items/any(property:property/shared+ne+null)` could be used to get a list of all drives that have shared drive items and expand the items. This is not allowed on the ms graph api.
What does work is `https://graph.microsoft.com/v1.0/me/drive/root/search(q='*')?filter=shared+ne+null` but it is slow as it relies on the search() function. I'd go with the semantically correct version.

interesting sidenote
* creating sharing links: https://learn.microsoft.com/en-us/graph/api/driveitem-createlink?view=graph-rest-1.0&tabs=http
* there is a [`/shares/{share-id}` endpoint](https://learn.microsoft.com/en-us/graph/api/resources/shareddriveitem?view=graph-rest-1.0) that can be used to directly access a shared drive item ... similar to `/drives/{drive-id}` ... 🤔

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -3094,19 +3094,23 @@ components:
     sharePointIdentitySet:
       type: object
       description: This resource is used to represent a set of identities associated with various events for an item, such as created by or last modified by.
-      properties:     
-        #application:
-        #  $ref: '#/components/schemas/identity'
-        #  description: Optional. The application associated with this action.
-        #device:
-        #  $ref: '#/components/schemas/identity'
-        #  description: Optional. The device associated with this action.
+      properties:
         user:
           $ref: '#/components/schemas/identity'
           description: Optional. The user associated with this action.
         group:
           $ref: '#/components/schemas/identity'
           description: Optional. The group associated with this action.
+
+        # unused  
+        #application:
+        #  $ref: '#/components/schemas/identity'
+        #  description: Optional. The application associated with this action.
+
+        # unused
+        #device:
+        #  $ref: '#/components/schemas/identity'
+        #  description: Optional. The device associated with this action.
         # other properties omitted for now
     identity:
       type: object
@@ -3161,12 +3165,18 @@ components:
         '@libre.graph.displayName':
           type: string
           description: Provides a user-visible display name of the link. Optional. Libregraph only.
+
+        # unused
         #webHtml:
         #  type: string
         #  description: For `embed` links, this property contains the HTML code for an `<iframe>` element that will embed the item in a webpage.
+
+        # unused
         #application:
         #  description: The app the link is associated with.
         #  $ref: '#/components/schemas/identity'
+
+        # unused
         #scope:
         #  type: string
         #  enum: [anonymous, organization, existingAccess, users]
@@ -3393,24 +3403,24 @@ components:
         displayName:
           type: string
           description: Provides a user-visible display name of the link. Optional. Libregraph only.
-        # to invite guests or specify which users should be sent the link we could add this
+        # unused. we could add thisto invite guests or specify which users should be sent the link
         #recipients:
         #  type: array
         #  items:
         #    $ref: '#/components/schemas/driveRecipient'
         #  description: A collection of recipients who will receive access and the sharing link.
 
-        # can be added later
+        # unused. can be added later
         #scope:
         #  type: string
         #  description: Optional. The scope of link to create. Either `anonymous`, `organization`, or `users`.
 
-        # In ocis recipients configure if they want to be notified
+        # unused. In ocis recipients configure if they want to be notified
         #sendNotification:
         #  type: boolean
         #  description: Optional. If `true`, this method sends a sharing link in an email to users specified in recipients. The default value is `false`. Optional.
 
-        # not supported by ocis
+        # unused. not supported by ocis
         #retainInheritedPermissions:
         #  type: boolean
         #  description: Optional. If `true` (default), any existing inherited permissions are retained on the shared item when sharing this item for the first time. If `false`, all existing permissions are removed when sharing for the first time.
@@ -3579,7 +3589,7 @@ components:
           description: "The drive alias can be used in clients to make the urls user friendly. Example: 'personal/einstein'. This will be used to resolve to the correct driveID."
         #driveType:
         # wie kommt man zum drive? web möchte im ui den pfad anzeigen, der wird aus dem driveAlias gebaut, aber die brauchen auch den driveType um das datenmodell konsistent zu haben
-        # momentan machen sie einen ocs list shares request und sichen sich dann den mountpoint anhand der id raus, aber cern möchte IMMER den personal space dazu sehen, incl pfad
+        # momentan machen sie einen ocs list shares request und sichen sich dann den mountpoint anhand der id raus, aber einige möchten IMMER den personal space dazu sehen, incl pfad
         path:
           type: string
           description: "The relative path of the item in relation to its drive root."
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -282,32 +282,21 @@ paths:
       summary: Create a sharing link for a DriveItem
       operationId: CreateLink
       description: |
-        You can use createLink action to share a driveItem via a sharing link.
+        You can use the createLink action to share a driveItem via a sharing link.
 
-        The **createLink** action will create a new sharing link if the specified link type doesn't already exist
-        for the calling application. If a sharing link of the specified type already exists for the app, the
-        existing sharing link will be returned.
+        The response will be a permission object with the link facet containing the created link details.
 
-        DriveItem resources inherit sharing permissions from their ancestors.
-      
         ## Link types
 
         For now, The following values are allowed for the type parameter.
         
-        | Type value      | Description                                                                                                                                                                                                           |
-        | --------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-        | view            | Creates a read-only link to the driveItem.                                                                                                                                                                            |
-        | edit            | Creates an read-write link to the driveItem.                                                                                                                                                                          |
-        | blocksDownload  | Creates a read-only link that blocks download to the driveItem.                                                                                                                                                       |
-        | createOnly      | Creates an upload-only link to the driveItem.                                                                                                                                                                         |
-
-        More built in types might be added with a speaking value but clients **MUST** treat the value as an
-        opaque identifier as custom types will be using UUIDs.
-
-      #  | embed           | Creates an embeddable link to the driveItem.                                                                                                                                                                          |
-      #  | review          | Creates a review link to the driveItem.                                                                                                                                                                               |
-      #  | addressBar      | Creates the default link that is shown in the browser address bars for newly created files. The organization admin configures whether this link type is supported, and what features are supported by this link type. |
-      #  | adminDefault    | Creates the default link to the driveItem as determined by the administrator of the organization. The policy is enforced for the organization by the admin                                                            |
+        | Value          | Display name      | Description                                                     |
+        | -------------- | ----------------- | --------------------------------------------------------------- |
+        | view           | View              | Creates a read-only link to the driveItem.                      |
+        | upload         | Upload            | Creates a read-write link to the folder driveItem.              |
+        | edit           | Edit              | Creates a read-write link to the driveItem.                     |
+        | createOnly     | File Drop         | Creates an upload-only link to the folder driveItem.            |
+        | blocksDownload | Secure View       | Creates a read-only link that blocks download to the driveItem. |
 
       parameters:
         - name: drive-id
@@ -329,44 +318,15 @@ paths:
         content:
           application/json:
             schema:
-              type: object
-              properties:
-                type:
-                  type: string
-                  description: Optional. The type of sharing link to create.
-                expirationDateTime:
-                  type: string
-                  description: Optional. A String with format of yyyy-MM-ddTHH:mm:ssZ of DateTime indicates the expiration time of the permission.
-                  format: date-time
-                password: 
-                  type: string
-                  description: Optional.The password of the sharing link that is set by the creator.
-
-                # to invite guests or specify which users should be sent the link we could add this
-                #recipients:
-                #  type: array
-                #  items:
-                #    $ref: '#/components/schemas/driveRecipient'
-                #  description: A collection of recipients who will receive access and the sharing link.
-
-                # can be added later
-                #scope:
-                #  type: string
-                #  description: Optional. The scope of link to create. Either `anonymous`, `organization`, or `users`.
-
-                # In ocis recipients configure if they want to be notified
-                #sendNotification:
-                #  type: boolean
-                #  description: Optional. If `true`, this method sends a sharing link in an email to users specified in recipients. The default value is `false`. Optional.
-
-                # not supported by ocis
-                #retainInheritedPermissions:
-                #  type: boolean
-                #  description: Optional. If `true` (default), any existing inherited permissions are retained on the shared item when sharing this item for the first time. If `false`, all existing permissions are removed when sharing for the first time.
+              $ref: '#/components/schemas/driveItemCreateLink'
             examples:
                 create viewer link:
                   value:
                     type: "view"
+                create editor link with custom display name:
+                  value:
+                    type: "edit"
+                    displayName: "Homework"
                 create editor link with expiry:
                   value:
                     type: "edit"
@@ -379,19 +339,27 @@ paths:
               schema:
                 $ref: '#/components/schemas/permission'
               examples:
-                send viewer invite:
+                create viewer link:
                   value:
                     - id: "7181275c-557e-4adf-8e44-abdcf5389b6a"
                       link:
                         type: "view"
                         webUrl: "https://cloud.example.org/s/YlaySZdHCRHUDeE"
-                send editor invite with expiry:
+                create editor link with custom display name:
+                  value:
+                    - id: "7181275c-557e-4adf-8e44-abdcf5389b6a"
+                      link:
+                        type: "edit"
+                        displayName: "Homework"
+                        webUrl: "https://cloud.example.org/s/fdleMyGeerkIZTJ"
+                create editor link with expiry:
                   value:
                     - id: "a373c812-d786-42ae-a65f-ed87ee4abd49"
                       link:
                         type: "edit"
                         webUrl: "https://cloud.example.org/s/MlnTDkhcBlihNzO"
         '207':
+        # FIXME describe partial success response
           description: Partial success response TODO
         default:
           $ref: '#/components/responses/error'
@@ -406,6 +374,8 @@ paths:
       description: |
         Sends a sharing invitation for a `driveItem`. A sharing invitation provides permissions to the
         recipients and optionally sends them an email with a sharing link.
+
+        The response will be a permission object with the grantedToV2 property containing the created grant details.
 
         ## Roles property values
         For now, roles are only identified by a uuid. There are no hardcoded aliases like `read` or `write` because role actions can be completely customized.
@@ -467,7 +437,7 @@ paths:
                   description: "Sharing with a groups requires setting the @libre.graph.recipient.type annotation."
                   value:
                     recipients:
-                    - "@libre.graph.recipient.type": group
+                    - '@libre.graph.recipient.type': group
                       objectId: "167cbee2-0518-455a-bfb2-031fe0621e5d"
                     roles: [ "dfa16e02-3d88-45d8-8894-5b33a7df637" ]
                 send invite to user granting custom actions:
@@ -475,7 +445,7 @@ paths:
                   value:
                     recipients:
                     - objectId: "4c510ada-c86b-4815-8820-42cdf82c3d51"
-                    actions: [ "libre.graph/driveItem/basic/read", "libre.graph/driveItem/path/update" ]
+                    '@libre.graph.permissions.actions': [ "libre.graph/driveItem/basic/read", "libre.graph/driveItem/path/update" ]
       responses:
         '200':
           description: Response
@@ -487,15 +457,7 @@ paths:
                 send viewer invite:
                   value:
                     - id: "34646ab6-be32-43c9-89e6-987e0c237e9b"
-                      roles: [ "read" ]
-                      grantedToV2:
-                        - user:
-                            id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
-                            displayName: "Albert Einstein"
-                send editor invite with expiry:
-                  value:
-                    - id: "81d5bad3-3eff-410a-a2ea-eda2d14d4474"
-                      roles: [ "write" ]
+                      roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
                       grantedToV2:
                         - user:
                             id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
@@ -504,26 +466,44 @@ paths:
                   description: multiple permissions are returned as each has received a dedicated sharing id
                   value:
                     - id: "81d5bad3-3eff-410a-a2ea-eda2d14d4474"
-                      roles: [ "write" ]
+                      roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
                       grantedToV2:
                         - user:
                             id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
                             displayName: "Albert Einstein"
                     - id: "b470677e-a7f5-4304-8ef5-f5056a21fff1"
-                      roles: [ "write" ]
+                      roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
                       grantedToV2:
                         - user:
                             id: "f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c"
                             displayName: "Marie Skłodowska Curie"
+                send editor invite with expiry:
+                  value:
+                    - id: "453b02be-4ec2-4e7d-b576-09fc153de812"
+                      roles: [ "a1c6f73e-482e-4078-a629-bbecb205abb" ]
+                      grantedToV2:
+                        - user:
+                            id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                            displayName: "Albert Einstein"
+                      expirationDateTime: "2018-07-15T14:00:00.000Z"
                 send manager invite to group:
                   value:
-                    - id: "81d5bad3-3eff-410a-a2ea-eda2d14d4474"
-                      roles: [ "owner" ]
+                    - id: "86765c0d-3905-444a-9b07-76201f8cf7df"
+                      roles: [ "dfa16e02-3d88-45d8-8894-5b33a7df637" ]
                       grantedToV2:
                         - group:
                             id: "167cbee2-0518-455a-bfb2-031fe0621e5d"
                             displayName: "Philosophy Haters"
+                send invite to user granting custom actions:
+                  value:
+                    - id: "c42b5cbd-2d65-42cf-b0b6-fb6d2b762256"
+                      grantedToV2:
+                        - user:
+                            id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                            displayName: "Albert Einstein"
+                      '@libre.graph.permissions.actions': [ "libre.graph/driveItem/basic/read", "libre.graph/driveItem/path/update" ]
         '207':
+        # FIXME describe partial success response
           description: Partial success response TODO
         '400':
           description: Bad request
@@ -558,7 +538,25 @@ paths:
 
         * For the owner of the item, all sharing permissions will be returned. This includes co-owners.
         * For a non-owner caller, only the sharing permissions that apply to the caller are returned.
-        * Sharing permission properties that contain secrets (e.g. `shareId` and `webUrl`) are only returned for callers that are able to create the sharing permission.
+        * Sharing permission properties that contain secrets (e.g. `webUrl`) are only returned for callers that are able to create the sharing permission.
+
+        All permission objects have an `id`. A permission representing
+        * a link has the `link` facet filled with details. 
+        * a share has the `roles` property set and the `grantedToV2` property filled with the grant recipient details.
+
+      # TODO hm, the ms graph api seems to always fill the `roles` property to indicate the actions granted by the link.
+      # | Value          | Display name      | Description                                                     | role?          |
+      # | -------------- | ----------------- | --------------------------------------------------------------- | -------------- |
+      # | view           | View              | Creates a read-only link to the driveItem.                      | viewer         |
+      # | upload         | Upload            | Creates a read-write link to the folder driveItem.              | uploader       |
+      # | edit           | Edit              | Creates a read-write link to the driveItem.                     | editor         |
+      # | createOnly     | File Drop         | Creates an upload-only link to the folder driveItem.            | createOnly     |
+      # | blocksDownload | Secure View       | Creates a read-only link that blocks download to the driveItem. | blocksDownload |
+      # AFAICT we wcill have to define the actions granted by a link type when we start to implement this
+      # but e.g. a createOnly role and blocksDownload should not show up in the drop down for sharing files internally ... 
+      # when we need to hardcode this anyway ... we should indicate that by using the buildIn flag ... but how do we prevent some roles from showing up in the internal sharing dialog ...
+      # for now leave out the roles property and hardcode the actions allowed by the link type
+
       parameters:
         - name: drive-id
           in: path
@@ -672,25 +670,47 @@ paths:
 
                 value:
                   - id: "67445fde-a647-4dd4-b015-fc5dafd2821d"
-                    roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
                     link:
                       type: "view"
                       webUrl: "https://cloud.example.org/s/fhGBMIkKFEHWysj"
-                  - id: "81d5bad3-3eff-410a-a2ea-eda2d14d4474"
-                    roles: [ "a1c6f73e-482e-4078-a629-bbecb205abb" ]
-                    expirationDateTime: "2018-07-15T14:00:00.000Z"
+
+                  - id: "34646ab6-be32-43c9-89e6-987e0c237e9b"
+                    roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
                     grantedToV2:
                       - user:
                           id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
                           displayName: "Albert Einstein"
-                    link:
-                      webUrl: "https://cloud.example.org/s/kXJpiJmMnQxVlaT"
-                  - id: "dda0bb77-ffe7-4867-a2bf-28af65b46aa3"
-                    actions: [ "libre.graph/driveItem/basic/read", "libre.graph/driveItem/path/update" ]
+                  - id: "81d5bad3-3eff-410a-a2ea-eda2d14d4474"
+                    roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
                     grantedToV2:
                       - user:
-                          id: "058bff95-6708-4fe5-91e4-9ea3d377588b"
-                          displayName: "Maurice Moss"
+                          id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                          displayName: "Albert Einstein"
+                  - id: "b470677e-a7f5-4304-8ef5-f5056a21fff1"
+                    roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
+                    grantedToV2:
+                      - user:
+                          id: "f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c"
+                          displayName: "Marie Skłodowska Curie"
+                  - id: "453b02be-4ec2-4e7d-b576-09fc153de812"
+                    roles: [ "a1c6f73e-482e-4078-a629-bbecb205abb" ]
+                    grantedToV2:
+                      - user:
+                          id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                          displayName: "Albert Einstein"
+                    expirationDateTime: "2018-07-15T14:00:00.000Z"
+                  - id: "86765c0d-3905-444a-9b07-76201f8cf7df"
+                    roles: [ "dfa16e02-3d88-45d8-8894-5b33a7df637" ]
+                    grantedToV2:
+                      - group:
+                          id: "167cbee2-0518-455a-bfb2-031fe0621e5d"
+                          displayName: "Philosophy Haters"
+                  - id: "c42b5cbd-2d65-42cf-b0b6-fb6d2b762256"
+                    grantedToV2:
+                      - user:
+                          id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                          displayName: "Albert Einstein"
+                    '@libre.graph.permissions.actions': [ "libre.graph/driveItem/basic/read", "libre.graph/driveItem/path/update" ]
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
@@ -3121,22 +3141,6 @@ components:
           type: string
           writeOnly: true
           description: The user's password
-    sharingInvitation:
-      description: The `SharingInvitation` resource groups invitation-related data items into a single structure.
-      type: object
-      properties:
-        email:
-          type: string
-          description: The email address provided for the recipient of the sharing invitation.
-          readOnly: true
-        invitedBy:
-          $ref: '#/components/schemas/identitySet'
-          description: Provides information about who sent the invitation that created this permission, if that information is available.
-          readOnly: true
-        signInRequired:
-          type: boolean
-          description: If `true` the recipient of the invitation needs to sign in in order to access the shared item.
-          readOnly: true
     sharingLink:
       description: |
         The `SharingLink` resource groups link-related data items into a single structure.
@@ -3144,47 +3148,36 @@ components:
         If a `permission` resource has a non-null `sharingLink` facet, the permission represents a sharing link (as opposed to permissions granted to a person or group).
       type: object
       properties:
-        application:
-          $ref: '#/components/schemas/identity'
-          description: The app the link is associated with.
         type:
-          type: string
-          enum: [view, edit, blocksDownload, createOnly]
-          description: |
-            The type of the link created.
-
-            | Value          | Description                                                                                                                                                                                                           |
-            | -------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-            | view           | Creates a read-only link to the driveItem.                                                                                                                                                                            |
-            | edit           | Creates an read-write link to the driveItem.                                                                                                                                                                          |
-            | blocksDownload | Creates a read-only link that blocks download to the driveItem.                                                                                                                                                       |
-            | createOnly     | Creates an upload-only link to the driveItem.                                                                                                                                                                         |
-          # there are more link types available in https://learn.microsoft.com/en-us/graph/api/driveitem-createlink?view=graph-rest-beta&tabs=http#link-types
-          #  | review         | Creates a review link to the driveItem.                                                                                                                                                                               |
-          #  | embed          | Creates an embeddable link to the driveItem.                                                                                                                                                                          |
-          #  | addressBar     | Creates the default link that is shown in the browser address bars for newly created files. The organization admin configures whether this link type is supported, and what features are supported by this link type. |
-          #  | adminDefault   | Creates the default link to the driveItem as determined by the administrator of the organization. The policy is enforced for the organization by the admin                                                            |            
-        scope:
-          type: string
-          enum: [anonymous, organization, existingAccess, users]
-          description: |
-            The scope of the link represented by this permission. Value `anonymous` indicates the link is usable by anyone, `organization` indicates the link is only usable for users signed into the same tenant.
-
-            | Value          | Description                                                                                                           |
-            | -------------- | --------------------------------------------------------------------------------------------------------------------- |
-            | anonymous      | Anyone with the link has access, without needing to sign in. This may include people outside of your organization.    |
-            | organization   | Anyone signed into your organization (tenant) can use the link to get access.                                         |
-            | existingAccess | Only people who have already been granted access to the item through other means can access the item using this link. |
-            | users          | The link grants access only to a specific list of people.                                                             |
+          $ref: '#/components/schemas/sharingLinkType'
         preventsDownload:
           type: boolean
           description: If `true` then the user can only use this link to view the item on the web, and cannot use it to download the contents of the item.
-        webHtml:
-          type: string
-          description: For `embed` links, this property contains the HTML code for an `<iframe>` element that will embed the item in a webpage.
+          readOnly: true
         webUrl:
           type: string
           description: A URL that opens the item in the browser on the website.
+          readOnly: true
+        '@libre.graph.displayName':
+          type: string
+          description: Provides a user-visible display name of the link. Optional. Libregraph only.
+        #webHtml:
+        #  type: string
+        #  description: For `embed` links, this property contains the HTML code for an `<iframe>` element that will embed the item in a webpage.
+        #application:
+        #  description: The app the link is associated with.
+        #  $ref: '#/components/schemas/identity'
+        #scope:
+        #  type: string
+        #  enum: [anonymous, organization, existingAccess, users]
+        #  description: |
+        #    The scope of the link represented by this permission. Value `anonymous` indicates the link is usable by anyone, `organization` indicates the link is only usable for users signed into the same tenant.
+        #    | Value          | Description                                                                                                           |
+        #    | -------------- | --------------------------------------------------------------------------------------------------------------------- |
+        #    | anonymous      | Anyone with the link has access, without needing to sign in. This may include people outside of your organization.    |
+        #    | organization   | Anyone signed into your organization (tenant) can use the link to get access.                                         |
+        #    | existingAccess | Only people who have already been granted access to the item through other means can access the item using this link. |
+        #    | users          | The link grants access only to a specific list of people.                                                             |
     user:
       description: Represents an Active Directory user object.
       type: object
@@ -3366,6 +3359,60 @@ components:
             $ref: '#/components/schemas/permission'
           description: The set of permissions for the item. Read-only. Nullable.
           readOnly: true
+    sharingLinkType:      
+        type: string
+        enum: [view, upload, edit, createOnly, blocksDownload]
+        description: |
+          The type of the link created.
+
+          | Value          | Display name      | Description                                                     |
+          | -------------- | ----------------- | --------------------------------------------------------------- |
+          | view           | View              | Creates a read-only link to the driveItem.                      |
+          | upload         | Upload            | Creates a read-write link to the folder driveItem.              |
+          | edit           | Edit              | Creates a read-write link to the driveItem.                     |
+          | createOnly     | File Drop         | Creates an upload-only link to the folder driveItem.            |
+          | blocksDownload | Secure View       | Creates a read-only link that blocks download to the driveItem. |
+        # there are more link types available in https://learn.microsoft.com/en-us/graph/api/driveitem-createlink?view=graph-rest-beta&tabs=http#link-types
+        #  | review         | Creates a review link to the driveItem.                                                                                                                                                                               |
+        #  | embed          | Creates an embeddable link to the driveItem.                                                                                                                                                                          |
+        #  | addressBar     | Creates the default link that is shown in the browser address bars for newly created files. The organization admin configures whether this link type is supported, and what features are supported by this link type. |
+        #  | adminDefault   | Creates the default link to the driveItem as determined by the administrator of the organization. The policy is enforced for the organization by the admin                                                            |
+    driveItemCreateLink:
+      type: object
+      properties:
+        type:
+          $ref: '#/components/schemas/sharingLinkType'
+        expirationDateTime:
+          type: string
+          description: Optional. A String with format of yyyy-MM-ddTHH:mm:ssZ of DateTime indicates the expiration time of the permission.
+          format: date-time
+        password: 
+          type: string
+          description: Optional.The password of the sharing link that is set by the creator.
+        displayName:
+          type: string
+          description: Provides a user-visible display name of the link. Optional. Libregraph only.
+        # to invite guests or specify which users should be sent the link we could add this
+        #recipients:
+        #  type: array
+        #  items:
+        #    $ref: '#/components/schemas/driveRecipient'
+        #  description: A collection of recipients who will receive access and the sharing link.
+
+        # can be added later
+        #scope:
+        #  type: string
+        #  description: Optional. The scope of link to create. Either `anonymous`, `organization`, or `users`.
+
+        # In ocis recipients configure if they want to be notified
+        #sendNotification:
+        #  type: boolean
+        #  description: Optional. If `true`, this method sends a sharing link in an email to users specified in recipients. The default value is `false`. Optional.
+
+        # not supported by ocis
+        #retainInheritedPermissions:
+        #  type: boolean
+        #  description: Optional. If `true` (default), any existing inherited permissions are retained on the shared item when sharing this item for the first time. If `false`, all existing permissions are removed when sharing for the first time.
     driveItemInvite:
       type: object
       properties:
@@ -3379,7 +3426,7 @@ components:
           items:
             type: string
           description: Specifies the roles that are to be granted to the recipients of the sharing invitation.
-        actions:
+        '@libre.graph.permissions.actions':
           type: array
           items:
             type: string
@@ -3719,11 +3766,13 @@ components:
         id:
           type: string
           description: The unique identifier of the permission among all permissions on the item. Read-only.
+          readOnly: true
         hasPassword:
           type: boolean
           description: |
             Indicates whether the password is set for this permission. This property only
             appears in the response. Optional. Read-only.
+          readOnly: true
         expirationDateTime:
           type: string
           description: An optional expiration date which limits the permission in time.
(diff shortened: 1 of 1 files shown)


butonic:
The web ui currently needs to PROPFIND all path segments to receive sharing information and look up the fileid for path segments... Some thoughts:

## how do we get all driveItems that between the root and a child /a/b/c/d.txt?
- GET all driveItems whose path is a prefix of a given path
- GET all driveItems whose path is a prefix of eg /a/b/c/d.txt
- GET /drive/items/?$filter=startswith('/a/b/c/d.txt',path)
- GET /drive/items/?$filter=startswith('/drive/root:/a/b/c/d.txt',path)

for that the path in the parentReference needs to have a determinable path:
```
            "parentReference": {
                "driveId": "c12644a14b0a7750",
                "driveType": "personal",
                "id": "C12644A14B0A7750!6227",
                "name": "linux-5.19-rc6",
                "path": "/drive/root:/linux/linux-5.19-rc6"
            },   
```
https://graph.microsoft.com/v1.0/drives/C12644A14B0A7750/items

better to always return the `/drives` endpoint with the drive id:

This should always work: `GET /drives/{uuid}/items/?$filter=startswith('/drives/{uuid}/a/b/c/d.txt',path)`

butonic:
I think sharing roles are best sent as a [instance annotation](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_InstanceAnnotations) of the collection returned when listing permissions:
```json
{
  "@libre.graph.permissions.roles.allowedValues": { "read", "write", "owner", "editor" },
  "value": [
    {
      "id": "67445fde-a647-4dd4-b015-fc5dafd2821d",
      "roles": [
        "read"
      ],
      "shareId": "TODO the share id?",
      "link": {
        "type": "view",
        "webUrl": "https://cloud.example.org/TODO-URL-to-use-in-the-browser-bar"
      }
    },
    {
      "id": "81d5bad3-3eff-410a-a2ea-eda2d14d4474",
      "roles": [
        "write"
      ],
      "shareId": "TODO the share id?",
      "expirationDateTime": "2018-07-15T14:00:00.000Z",
      "hasPassword": "false,",
      "grantedTo": [
        {
          "user": {
            "id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
            "displayName": "Albert Einstein"
          }
        }
      ],
      "invitation": {
        "email": "einstein@example.org",
        "signInRequired": true
      },
      "link": {
        "webUrl": "https://cloud.example.org/TODO-URL-to-use-in-the-browser-bar"
      }
    }
  ]
}
```

The allowed values can be configured in a way, similar to other role and permission relationships are managed in microsoft graph: the [`roleManagement` resource](https://learn.microsoft.com/en-us/graph/api/resources/rolemanagement?view=graph-rest-1.0#relationships).

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,360 @@ paths:
...
+                  type: string
+                  description: A plain text formatted message that is included in the sharing invitation.
+                  maxLength: 2000
+                requireSignIn:
+                  type: boolean
+                  description: Specifies whether the recipient of the invitation is required to sign-in to view the shared item.
+                sendInvitation:
+                  type: boolean
+                  description: If `true`, a sharing link is sent to the recipient. Otherwise, a permission is granted directly without sending a notification.
+                roles:
```
micbar, line 316: Should be role IDs. Ok to keep it as string here, but in our case we would use uuids

micbar: But we should make it an array of strings.

api/openapi-spec/v1.0.yaml
```diff
@@ -2738,21 +3307,152 @@ components:
...
+          type: boolean
+        # leaving this out as it is already deprecated in msgraph and I do not think we need it right now
+        #resourceScopes:
+        #  type: string
+        #  description: |
+        #    List of the scopes or permissions the role definition applies to. Currently only / is supported.
+        #    Read-only when isBuiltIn is true. DO NOT USE. This will be deprecated soon. Attach scope to role assignment.
+        rolePermissions:
+          description: List of permissions included in the role. Read-only when **isBuiltIn** is `true`.
+          type: array
```
micbar, line 3375: Should we use / expose the CS3 ResourcePermissions directly? I think there speaks nothing against that.

api/openapi-spec/v1.0.yaml
rhafer: Yeah, the MS Graph docs describe this to be a `Collection(String)`. So this should be:

```
roles:
  type: array
  items:
    type: string
```


Though I guess for ocis this will always just be a single value?

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,360 @@ paths:
...
+                  type: string
+                  description: The password set on the invite by the creator. Optional .
+                #retainInheritedPermissions:
+                #  type: boolean
+                #  description: Optional. If `true` (default), any existing inherited permissions are retained on the shared item when sharing this item for the first time. If `false`, all existing permissions are removed when sharing for the first time.
+            examples:
+                create write share:
+                  value:
+                    recipients:
+                    - email: "einstein@example.org"
```
rhafer, line 338: Can we make this example using the `objectId` instead of `email`? I guess, at least in the beginning, `objectId` is to going to be property used in most cases.

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,360 @@ paths:
...
+                #sendInvitation:
+                #  type: boolean
+                #  description: If `true`, a sharing link is sent to the recipient. Otherwise, a permission is granted directly without sending a notification.
+                roles:
+                  type: string
+                  description: Specifies the roles that are to be granted to the recipients of the sharing invitation.
+                  # TODO define which roles are available
+                expirationDateTime:
+                  type: string
+                  description: Specifies the dateTime after which the permission expires. Only applicable for `sharingLink` permissions.
```
rhafer, line 326: This is not true for ocis, is it? We can set an expiration date on any share there.

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,360 @@ paths:
...
+            type: string
+          x-ms-docs-key-type: item
+        - name: perm-id
+          in: path
+          description: 'key: id of permission'
+          required: true
+          schema:
+            type: string
+          x-ms-docs-key-type: permission
+        - name: If-Match
```
rhafer, line 541: Does this need to be part of the first implementation? If not, I'd suggest to remove/comment it from the spec for now to not create wrong expectations. (Same for the other occurrences of `If-Match` and similar Headers. 

(Sidenote: I think we don't have any support for `If-Match` headers in the graph service yet.

api/openapi-spec/v1.0.yaml
```diff
@@ -639,6 +993,56 @@ paths:
...
+                title: Collection of driveItems
+                type: object
+                properties:
+                  value:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/driveItem'
+                    maxItems: 20
+                  '@odata.nextLink':
+                    type: string
```
rhafer, line 1020: Is paging a requirement for the first iteration of this API?

rhafer (changes requested):
In general I think we need some good examples that fit the use-cases for ocis. 

It's still somewhat unclear to me how the `/invite` this is supposed to work.
Is it only supposed to be used for sharing with other users and groups? Are "public" shares supposed to be created via `/createLink`

If that is the case we might want to remove the `requireSignIn` and `password` properties from  the properites of the `invite` body.

api/openapi-spec/v1.0.yaml
butonic: I think we should use strings for built in roles. Custom roles will use UUIDs.

api/openapi-spec/v1.0.yaml
butonic: For now @jkoberg and I decided to not expose role permissions. The MS Graph way of describing role permissions is extensive but for now we can omit that because the web ui only needs to have an endpoint that we can use to list the existing roles with their id, a short name and a description.

api/openapi-spec/v1.0.yaml
butonic: changed to array

api/openapi-spec/v1.0.yaml
butonic: commented for now

api/openapi-spec/v1.0.yaml
butonic: correct ocis can expire user shares

butonic:
@jkoberg @micbar @rhafer I updated the examples and added the sharedByMe endpoint. please rereview

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,516 @@ paths:
...
+      requestBody:
+        description: In the request body, provide a JSON object with the following parameters.
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                type:
+                  type: string
+                  description: Optional. The type of sharing link to create.
```
rhafer, line 330: If this is  optional, what kind of link is created, when no type is specified? If only a certain set of types is allowed I think we should specify those values here via `enum`

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,516 @@ paths:
...
+                #  description: Optional. If `true`, this method sends a sharing link in an email to users specified in recipients. The default value is `false`. Optional.
+
+                # not supported by ocis
+                #retainInheritedPermissions:
+                #  type: boolean
+                #  description: Optional. If `true` (default), any existing inherited permissions are retained on the shared item when sharing this item for the first time. If `false`, all existing permissions are removed when sharing for the first time.
+            examples:
+                create viewer link:
+                  value:
+                    roles: [ "view" ]
```
rhafer, line 363: `roles` is not a valid property for the `createLink` body. I guess you meant `type` here? `type` would be a single value though (not an array)

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,516 @@ paths:
...
+                create editor link with expiry:
+                  value:
+                    roles: [ "edit" ]
```
rhafer, line 366: `type`?

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,516 @@ paths:
...
+                    - objectId: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                    roles: [ "read" ]
+                send editor invite with expiry:
+                  value:
+                    recipients:
+                    - objectId: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                    roles: [ "write" ]
+                    expirationDateTime: "2018-07-15T14:00:00.000Z"
+                send manager invite to group:
+                  description: "sharing with a groups needs to use the group name as alias"
```
rhafer, line 497: Why is that?

api/openapi-spec/v1.0.yaml
```diff
@@ -269,6 +269,516 @@ paths:
...
+            type: string
+          x-ms-docs-key-type: item
+        - name: perm-id
+          in: path
+          description: 'key: id of permission'
+          required: true
+          schema:
+            type: string
+          x-ms-docs-key-type: permission
+        - name: If-Match
```
rhafer, line 769: Yet another `If-Match` Header

rhafer (changes requested):

api/openapi-spec/v1.0.yaml
butonic: AFAICT the link type is redundant to sharing roles. I made the latter discoverable at the `/roleManagement/permissions/roleDefinitions` endpoint. I think we can probably leave out link types and reuse roles for links as well ...

api/openapi-spec/v1.0.yaml
```diff
@@ -639,6 +1146,110 @@ paths:
...
+                    lastModifiedDateTime: "2021-09-03T14:09:25.503Z"
+                    name: "March Proposal.docx"
+                    parentReference:
+                      driveId: "1991210caf"
+                      driveType: personal
+                    remoteItem:
+                      id: 02d7a7df-a6f7-44cc-848a-3c98f7dd5046
+                      name: "March Proposal.docx"
+                      size: 19121
+                    shared:
```
rhafer, line 1243: I think the indentation is wrong here. `shared` is a property of `remoteItem`, isn't it?

butonic:
interesting:
grafik

Removing' a folder only seems to wipe it from the lists that are queried by the web ui ...

butonic:
@dschmidt we were talking about the wish of the web client to list all shares in one request because the internal data model prefers that, or did I misunderstand that? Listing /sharedWithMe returns a list of driveItems. On Ms graph it only returns all shares, regardless of their mount status ... because it does not have a mount status. The Ms graph API represents mount points as driveItems with a remoteItem that wraps the shared driveItem. This way of wrapping the share in the mount point prevents listing shares without a mount point .... hmmm ... /sharedWithMe is a list of driveItems, so we could list unmounted shares as drive items without wrapping them in a drive Item that represents the mountpoint. While the graph API does not do that I think it semantically fits the concept. I'll add an example tomorrow.

Now about hiding shares... we could use a `@libregraph.hidden` annotation.

dschmidt:
> @dschmidt we were talking about the wish of the web client to list all shares in one request because the internal data model prefers that, or did I misunderstand that?

Yeah, from Web's POV "sharedness"  is just a property of a Drive(Item). So for us it would be nice to retrieve all shares accepted or not from the same endpoint. 
Especially do we need the same data available, e.g. the shareId... and anything else we might need to access it/match urls

butonic:
@dschmidt yeah ... so .... let me clarify what the `/me/sharedWithMe` endpoint could return:
```json
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
  "value": [
    {
      "id": "u-u-id-of-mountpoint",
      "name": "November-December Ad Proposals.pptx",
      "size": 100984,
      "parentReference": {
        "driveType": "personal",
        "driveId": "u-u-id-of-drive-containing-the-pointpoint",
        "id": "parent-u-u-id"
      },
      "createdBy": { ... },
      "lastModifiedBy": { ... },
      "remoteItem": {
        "id": "u-u-id-of-shared-driveItem",
        "name": "November-December Ad Proposals.pptx",
        "size": 100984,
        "parentReference": {
          "driveType": "personal",
          "driveId": "u-u-id-of-drive-containing-the-item"
          // no "id" because recipient is not allowed to see parent
        },
        "shared": {
          "scope": "users",
          "sharedBy": {
            "user": { ... }
          }
        }
      }
    },
    {
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "size": 17776,
      "parentReference": {
        "driveType": "personal",
        "driveId": "u-u-id-of-drive-containing-the-item"
        // no "id" because recipient is not allowed to see parent
      },
      "shared": {
        "scope": "users",
        "sharedBy": {
          "user": { ... }
        }
      }
    }
  ]
}
```

ms graph only returns a superficial `share` object with few properties, but maybe sufficient to build a list in the ui. when selecting a share you would then fetch the share details ...

butonic:
with regards to a `@libre.graph.hidden` annotation: we can follow the [odata spec](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_InstanceAnnotations) and include libregraph specific properties as annotations.
```
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
  "value": [
    {
      "@libre.graph.hidden": true,
      "id": "408082ff-98ac-4b48-aa47-3ae99c4a4bb0",
      ...
    },
    {
      "@libre.graph.hidden": true,
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      ...
    }
  ]
}
```
and changing the annotation would be a PATCH request:
```
PATCH /drives/{driveid}/items/{itemid}
{
  "@libre.graph.hidden": true
}
```

butonic:
@micbar @rhafer  I rebased the sharing NG pr and tried to clean up the permission related endpoints
I added an actions property to allow granting custom pormissions (which sucks ... but ... whatever) ...
Unfortunately, we also need to think about the drives/{drive-id}/items/{item-id}/createLink endpoint
the ms graph api uses link 'types' to express what is possible with the link.
The actions that are granted are determined by the link type as well. So you could argue that 'type' is just another name for 'role' and we should reuse the actions and unified role permissions from the /drives/{drive-id}/items/{item-id}/invite endpoint. But then can we please add the builtin and enabled flags for unifiedRoleDefinitions?
Also we would have to use a different set of unifiedRoleDefinitions at eg. /roleManagement/files.links/roleDefinitions? then we should probably also use /roleManagement/files.roles/roleDefinitions

butonic:
@dschmidt 

> The remoteItem points to the origin of the share (?) - why would it be missing for a grant? I don't get that logic.

Here is a shared driveItem. It is just a driveItem representing the actual remote item aka granted resource:
Here is a shared driveItem. It is just a driveItem representing the actual remote item aka granted resource:
```json5
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
  "value": [
    {
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "parentReference": {
        "driveType": "personal",
        "driveId": "u-u-id-of-drive-containing-the-item"
        // no "id" because the recipient is not allowed to see parent
      },
    }
  ]
}
```

When the user 'mounts' that share (by POSTing a drive item in the share jail drive `a0ca6a90-a365-4782-871e-d44447bbc668`) it will be wrapped as a `remoteItem` in a drive item that represents the mountpoint. Initially it will have the same name but the user is free to change it.
```json5
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
  "value": [
    {
      "id": "a0ca6a90-a365-4782-871e-d44447bbc668!u-u-id-of-mountpoint",
      "name": "Marketing Term Successes International.xlsx",
      "parentReference": {
        "driveType": "personal",
        "driveId": "a0ca6a90-a365-4782-871e-d44447bbc668",
        "id": "a0ca6a90-a365-4782-871e-d44447bbc668!a0ca6a90-a365-4782-871e-d44447bbc668"
      },
      "remoteItem": {
        "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
        "name": "Marketing Term Successes International.xlsx",
        "parentReference": {
          "driveType": "personal",
          "driveId": "u-u-id-of-drive-containing-the-item"
          // no "id" because the recipient is still not allowed to see parent
        },
      }
    },
  ]
}
```


> Why can't we have a `@libregraph.accepted` annotation?

Because it is redundant and we try to stay as close to the ms graph api as possible.

> Ftr: The `remoteItems` we currently receive contain a path (yes the path in the original space), we want to get rid of this dependency, but for now we still depend on that for the full share owner paths feature.

The `parentReference` of a `remoteItem` is an [itemReference](https://learn.microsoft.com/en-us/graph/api/resources/itemreference?view=graph-rest-1.0#properties) that has a `path` property. While this does leak the path in the owners space we can use it and make exposing it configurable, IMO.

> Another reminder: we need access to the `shareId`, otherwise we cannot efficiently match urls to shares if we don't have access to full paths (from url and at some point neither from api) and if we may not walk up dirs to find the share, if we just have a fileId

- [ ] I'll bring in the shareId, it exists as a property on permissions and it correlatios with the ocs share id.
- [ ] I also still need to add the `@libre.graph.hidden` annotation to permissions
- [ ] I need to make clear that `/me/drive/sharedWithMe` will noet return the permissions, only `/me/drive/sharedWithMe?$expand=permissions` ... or whatever the correct notation for expanding the permissions of all drive items is will.
  - maybe a drive items `shared` property is good enough to indicate that an item has been shared: https://learn.microsoft.com/en-us/graph/api/resources/shared?view=graph-rest-beta
    ```
    {
      "owner": { "@odata.type": "microsoft.graph.identitySet" },
      "scope": "anonymous | organization | users",
      "sharedBy": { "@odata.type": "microsoft.graph.identitySet" },
      "sharedDateTime": "datetime"
    }
    ```

dschmidt:
> @dschmidt
> 
> > The remoteItem points to the origin of the share (?) - why would it be missing for a grant? I don't get that logic.
> 
> Here is a shared driveItem. It is just a driveItem representing the actual remote item aka granted resource: Here is a shared driveItem. It is just a driveItem representing the actual remote item aka granted resource:
> 
> ```js
> {
>   "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
>   "value": [
>     {
>       "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
>       "name": "Marketing Term Successes International.xlsx",
>       "parentReference": {
>         "driveType": "personal",
>         "driveId": "u-u-id-of-drive-containing-the-item"
>         // no "id" because the recipient is not allowed to see parent
>       },
>     }
>   ]
> }
> ```
> 
> When the user 'mounts' that share (by POSTing a drive item in the share jail drive `a0ca6a90-a365-4782-871e-d44447bbc668`) it will be wrapped as a `remoteItem` in a drive item that represents the mountpoint. Initially it will have the same name but the user is free to change it.
> 
> ```js
> {
>   "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
>   "value": [
>     {
>       "id": "a0ca6a90-a365-4782-871e-d44447bbc668!u-u-id-of-mountpoint",
>       "name": "Marketing Term Successes International.xlsx",
>       "parentReference": {
>         "driveType": "personal",
>         "driveId": "a0ca6a90-a365-4782-871e-d44447bbc668",
>         "id": "a0ca6a90-a365-4782-871e-d44447bbc668!a0ca6a90-a365-4782-871e-d44447bbc668"
>       },
>       "remoteItem": {
>         "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
>         "name": "Marketing Term Successes International.xlsx",
>         "parentReference": {
>           "driveType": "personal",
>           "driveId": "u-u-id-of-drive-containing-the-item"
>           // no "id" because the recipient is still not allowed to see parent
>         },
>       }
>     },
>   ]
> }
> ```
> [...]
> The `parentReference` of a `remoteItem` is an [itemReference](https://learn.microsoft.com/en-us/graph/api/resources/itemreference?view=graph-rest-1.0#properties) that has a `path` property. While this does leak the path in the owners space we can use it and make exposing it configurable, IMO.


I'm still utterly confused.

This is what we have right now? You say only the `remoteItem` contains a path and what we have right now is that the `remoteItem` has a path and it points to the origin of the share. Like here:

```
{
    "value": [
{
            "driveAlias": "mountpoint/b",
            "driveType": "mountpoint",
            "id": "a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668!1391713f-411d-4993-8692-3f0c4eb7b566:4c510ada-c86b-4815-8820-42cdf82c3d51:84d2d477-aa22-4a2f-bfb9-755b4f262b4b",
            "name": "b",
            "owner": {
                "user": {
                    "displayName": "",
                    "id": "4c510ada-c86b-4815-8820-42cdf82c3d51"
                }
            },
            "root": {
                "id": "a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668!1391713f-411d-4993-8692-3f0c4eb7b566:4c510ada-c86b-4815-8820-42cdf82c3d51:84d2d477-aa22-4a2f-bfb9-755b4f262b4b",
                "remoteItem": {
                    "driveAlias": "personal/einstein",
                    "eTag": "\"47ff78a6c30adbd742d075e87f67c0dc\"",
                    "folder": {},
                    "id": "1391713f-411d-4993-8692-3f0c4eb7b566$4c510ada-c86b-4815-8820-42cdf82c3d51!de53429a-df40-4153-b395-e7329dc8a6f0",
                    "lastModifiedDateTime": "2023-09-13T13:38:28.720585898Z",
                    "name": "b",
                    "path": "/a/b",
                    "rootId": "1391713f-411d-4993-8692-3f0c4eb7b566$4c510ada-c86b-4815-8820-42cdf82c3d51!4c510ada-c86b-4815-8820-42cdf82c3d51",
                    "size": 4596,
                    "webDavUrl": "https://host.docker.internal:9200/dav/spaces/1391713f-411d-4993-8692-3f0c4eb7b566$4c510ada-c86b-4815-8820-42cdf82c3d51%214c510ada-c86b-4815-8820-42cdf82c3d51/a/b"
                },
                "webDavUrl": "https://host.docker.internal:9200/dav/spaces/a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668%211391713f-411d-4993-8692-3f0c4eb7b566:4c510ada-c86b-4815-8820-42cdf82c3d51:84d2d477-aa22-4a2f-bfb9-755b4f262b4b"
            },
            "webUrl": "https://host.docker.internal:9200/f/a0ca6a90-a365-4782-871e-d44447bbc668$a0ca6a90-a365-4782-871e-d44447bbc668%211391713f-411d-4993-8692-3f0c4eb7b566:4c510ada-c86b-4815-8820-42cdf82c3d51:84d2d477-aa22-4a2f-bfb9-755b4f262b4b"
        }
]
```

Now if that would be the same for the endpoint you describe, how are we supposed to access that path for a not accepted share? That's what we've been talking about all the time...

butonic:
@dschmidt ok, let me add the `path` to a drive item representing a shared resource (aka not the mountpoint / a nod yet accepted share). Assuming someone from the marketing team shared the file `Marketing Term Successes International.xlsx` in the `/International` folder of the `marketing-space-id` space. Then the recipient could see the following response when listing the `/me/drive/sharedWithMe`. I omitted `size`, `recipient`, etc to focus on the `path`.
```json
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
  "value": [
    {
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "parentReference": {
        "driveType": "project",
        "driveId": "marketing-space-id",
        "path": "/drives/{driveid}/root:/International"
      },
    }
  ]
}
```

Now, let me add some more properties to the `remoteItem`. We could use a `shared` property that holds some information about the owner:
```json
{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
  "value": [
    {
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "parentReference": {
        "driveType": "project",
        "driveId": "marketing-space-id",
        "path": "/drives/{driveid}/root:/International"
      },
      "shared": {
        "sharedBy": {
          "user": {
            "displayName": "Albert Einstein",
            "id": "4c510ada-c86b-4815-8820-42cdf82c3d51"
          }
        },
        "sharedDateTime": "2023-09-22T12:38:12.2577389Z"
      }
    }
  ]
}
```

This does not yet contain the permissions, but it is probably good enough to build a ui that indicates who created the share.

If all permissions are needed we can implement expanding the permissions:
```json
GET /me/drive/sharedWithMe?$expand=permissions

{
  "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#Collection(driveItem)",
  "value": [
    {
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "parentReference": {
        "driveType": "project",
        "driveId": "marketing-space-id",
        "path": "/drives/{driveid}/root:/International"
      },
      "shared": {
        "sharedBy": {
          "user": {
            "displayName": "Albert Einstein",
            "id": "4c510ada-c86b-4815-8820-42cdf82c3d51"
          }
        },
        "sharedDateTime": "2023-09-22T12:38:12.2577389Z"
      },
      "permissions": [
        {
          "id": "5b96d01b-2ca6-4c48-a870-f5cc9aea435c", // this is the ocs share id
          "roles": [
            "a1c6f73e-482e-4078-a629-bbecb205abb" // this corresponds to a sharing role from the /roleManagement/permissions/roleDefinitions endpoint
          ],
          "grantedToV2": [
            {
              "user": {
                "displayName": "Marie Skłodowska Curie",
                "id": "f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c"
              }
            }
          ]
        }
      ]
    }
  ]
}
```

butonic:
When navigating into a random driveItem how can we find out the path up to the driveItem the user has been granted access to?

Every drilveItem has a parentReference. One of the propertios is `path` which we can use to return the whole path, relative to the drive root. See https://learn.microsoft.com/en-us/graph/api/driveitem-get?view=graph-rest-1.0&tabs=http
```json
GET /graph/v1.0/drives/{driveId}/items/{driveItemId}

{
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "parentReference": {
        "driveType": "project",
        "driveId": "marketing-space-id",
        "path": "/drives/{driveId}/root:/International/Foo/Bar"
      }
}
```

This should also works for shares by using a path that 'starts' at the shared resource
```json
GET /graph/v1.0/drives/{driveId}/items/{driveItemId}

{
      "id": "7fd82e03-09af-4b38-8d36-c7f7ec83cd99",
      "name": "Marketing Term Successes International.xlsx",
      "parentReference": {
        "driveType": "project",
        "driveId": "marketing-space-id",
        "path": "/drives/{driveId}/items/{driveItemId}:/International"
      }
}
```

A request that fetches all driveItems with name and id could be expressed as:
```
GET /drives/{driveId}/items/?$filter=startswith('/drives/{driveId}/root:/a/b/c/d.txt',path)
```

concrete example: We want to get all driveItems for `/a/b/c/d/e/f/myfile.txt`
```
GET /drives/{driveId}/items/?$filter=startswith('/drives/{driveId}/root:/a/b/c/d/e/f/myfile.txt', path)
```
`path` here is the `path` property of the driveItems. `startsWith` now selects all drive items whose path is a prefix of the files path.

butonic:
@micbar I think I have now added all needed properties. odata uses vocabularies to extend entities and properties. I am using the SAP UI vocabulary to add the `@UI.Hidden` flag users can toggle to hide unwanted received permissions. I also tried to be spec conforamnt and used `@libre.graph` as the namespace for all the other custom annotations, introducing the libre graph odata vocabulary. Now ... I have no idea what any code generators will make of this, but this is how the spec says we should extend an existing odata api.

butonic:
The `@libre.graph...` and `@UI.Hidden` annotations look fine:
c++
```c++

// ...

     QList<QString> libre_graph_permissions_actions;
     bool libre_graph_permissions_actions_isSet;
     bool libre_graph_permissions_actions_isValid;

     bool ui_hidden;
     bool ui_hidden_isSet;
     bool ui_hidden_isValid;

// ...

        d->libre_graph_permissions_actions_isSet = false;
        d->libre_graph_permissions_actions_isValid = false;

        d->ui_hidden_isSet = false;
        d->ui_hidden_isValid = false;

// ...

    d->libre_graph_permissions_actions_isValid = ::OpenAPI::fromJsonValue(d->libre_graph_permissions_actions, json[QString("@libre.graph.permissions.actions")]);
    d->libre_graph_permissions_actions_isSet = !json[QString("@libre.graph.permissions.actions")].isNull() && d->libre_graph_permissions_actions_isValid;

    d->ui_hidden_isValid = ::OpenAPI::fromJsonValue(d->ui_hidden, json[QString("@UI.Hidden")]);
    d->ui_hidden_isSet = !json[QString("@UI.Hidden")].isNull() && d->ui_hidden_isValid;

// ...

    if (d->libre_graph_permissions_actions.size() > 0) {
        obj.insert(QString("@libre.graph.permissions.actions"), ::OpenAPI::toJsonValue(d->libre_graph_permissions_actions));
    }
    if (d->ui_hidden_isSet) {
        obj.insert(QString("@UI.Hidden"), ::OpenAPI::toJsonValue(d->ui_hidden));
    }

// ...


QList<QString> OAIPermission::getLibreGraphPermissionsActions() const {
    Q_D(const OAIPermission);
    if(!d){
        return {};
    }
    return d->libre_graph_permissions_actions;
}
void OAIPermission::setLibreGraphPermissionsActions(const QList<QString> &libre_graph_permissions_actions) {
    Q_D(OAIPermission);
    Q_ASSERT(d);

    d->libre_graph_permissions_actions = libre_graph_permissions_actions;
    d->libre_graph_permissions_actions_isSet = true;
}

bool OAIPermission::is_libre_graph_permissions_actions_Set() const{
    Q_D(const OAIPermission);
    if(!d){
        return false;
    }

    return d->libre_graph_permissions_actions_isSet;
}

bool OAIPermission::is_libre_graph_permissions_actions_Valid() const{
    Q_D(const OAIPermission);
    if(!d){
        return false;
    }
    return d->libre_graph_permissions_actions_isValid;
}

bool OAIPermission::isUiHidden() const {
    Q_D(const OAIPermission);
    if(!d){
        return {};
    }
    return d->ui_hidden;
}
void OAIPermission::setUiHidden(const bool &ui_hidden) {
    Q_D(OAIPermission);
    Q_ASSERT(d);

    d->ui_hidden = ui_hidden;
    d->ui_hidden_isSet = true;
}

bool OAIPermission::is_ui_hidden_Set() const{
    Q_D(const OAIPermission);
    if(!d){
        return false;
    }

    return d->ui_hidden_isSet;
}

bool OAIPermission::is_ui_hidden_Valid() const{
    Q_D(const OAIPermission);
    if(!d){
        return false;
    }
    return d->ui_hidden_isValid;
}

// ...
```

go
```go
// ...

	LibreGraphPermissionsActions []string `json:"@libre.graph.permissions.actions,omitempty"`
	// Properties or facets (see UI.Facet) annotated with this term will not be rendered if the annotation evaluates to true. Users can set this to hide permissons.
	UIHidden *bool `json:"@UI.Hidden,omitempty"`

// ...


	if !IsNil(o.LibreGraphPermissionsActions) {
		toSerialize["@libre.graph.permissions.actions"] = o.LibreGraphPermissionsActions
	}
	if !IsNil(o.UIHidden) {
		toSerialize["@UI.Hidden"] = o.UIHidden
	}

// ...

// GetLibreGraphPermissionsActions returns the LibreGraphPermissionsActions field value if set, zero value otherwise.
func (o *Permission) GetLibreGraphPermissionsActions() []string {
	if o == nil || IsNil(o.LibreGraphPermissionsActions) {
		var ret []string
		return ret
	}
	return o.LibreGraphPermissionsActions
}

// GetLibreGraphPermissionsActionsOk returns a tuple with the LibreGraphPermissionsActions field value if set, nil otherwise
// and a boolean to check if the value has been set.
func (o *Permission) GetLibreGraphPermissionsActionsOk() ([]string, bool) {
	if o == nil || IsNil(o.LibreGraphPermissionsActions) {
		return nil, false
	}
	return o.LibreGraphPermissionsActions, true
}

// HasLibreGraphPermissionsActions returns a boolean if a field has been set.
func (o *Permission) HasLibreGraphPermissionsActions() bool {
	if o != nil && !IsNil(o.LibreGraphPermissionsActions) {
		return true
	}

	return false
}

// SetLibreGraphPermissionsActions gets a reference to the given []string and assigns it to the LibreGraphPermissionsActions field.
func (o *Permission) SetLibreGraphPermissionsActions(v []string) {
	o.LibreGraphPermissionsActions = v
}

// GetUIHidden returns the UIHidden field value if set, zero value otherwise.
func (o *Permission) GetUIHidden() bool {
	if o == nil || IsNil(o.UIHidden) {
		var ret bool
		return ret
	}
	return *o.UIHidden
}

// GetUIHiddenOk returns a tuple with the UIHidden field value if set, nil otherwise
// and a boolean to check if the value has been set.
func (o *Permission) GetUIHiddenOk() (*bool, bool) {
	if o == nil || IsNil(o.UIHidden) {
		return nil, false
	}
	return o.UIHidden, true
}

// HasUIHidden returns a boolean if a field has been set.
func (o *Permission) HasUIHidden() bool {
	if o != nil && !IsNil(o.UIHidden) {
		return true
	}

	return false
}

// SetUIHidden gets a reference to the given bool and assigns it to the UIHidden field.
func (o *Permission) SetUIHidden(v bool) {
	o.UIHidden = &v
}

// ...
```

typescript
```ts
// ...

    /**
     * Use this to create a permission with custom actions.
     * @type {Array<string>}
     * @memberof Permission
     */
    '@libre.graph.permissions.actions'?: Array<string>;
    /**
     * Properties or facets (see UI.Facet) annotated with this term will not be rendered if the annotation evaluates to true. Users can set this to hide permissons.
     * @type {boolean}
     * @memberof Permission
     */
    '@UI.Hidden'?: boolean;

// ...
```

PHP
```php
// ...

    /**
      * Array of property to type mappings. Used for (de)serialization
      *
      * @var string[]
      */
    protected static $openAPITypes = [
        //...
        'at_libre_graph_permissions_actions' => 'string[]',
        'at_ui_hidden' => 'bool'
    ];

// ...

    /**
      * Array of property to format mappings. Used for (de)serialization
      *
      * @var string[]
      * @phpstan-var array<string, string|null>
      * @psalm-var array<string, string|null>
      */
    protected static $openAPIFormats = [
        //...
        'at_libre_graph_permissions_actions' => null,
        'at_ui_hidden' => null
    ];

// ...

    /**
     * Array of attributes where the key is the local name,
     * and the value is the original name
     *
     * @var string[]
     */
    protected static $attributeMap = [
        //...
        'at_libre_graph_permissions_actions' => '@libre.graph.permissions.actions',
        'at_ui_hidden' => '@UI.Hidden'
    ];

// ...

    /**
     * Array of attributes to setter functions (for deserialization of responses)
     *
     * @var string[]
     */
    protected static $setters = [
        //...
        'at_libre_graph_permissions_actions' => 'setAtLibreGraphPermissionsActions',
        'at_ui_hidden' => 'setAtUiHidden'
    ];

// ...

    /**
     * Array of attributes to getter functions (for serialization of requests)
     *
     * @var string[]
     */
    protected static $getters = [
        //...
        'at_libre_graph_permissions_actions' => 'getAtLibreGraphPermissionsActions',
        'at_ui_hidden' => 'getAtUiHidden'
    ];

// ...

    /**
     * Constructor
     *
     * @param mixed[] $data Associated array of property values
     *                      initializing the model
     */
    public function __construct(array $data = null)
    {
        //...
        $this->setIfExists('at_libre_graph_permissions_actions', $data ?? [], null);
        $this->setIfExists('at_ui_hidden', $data ?? [], null);
    }

// ...

    /**
     * Gets at_libre_graph_permissions_actions
     *
     * @return string[]|null
     */
    public function getAtLibreGraphPermissionsActions()
    {
        return $this->container['at_libre_graph_permissions_actions'];
    }

    /**
     * Sets at_libre_graph_permissions_actions
     *
     * @param string[]|null $at_libre_graph_permissions_actions Use this to create a permission with custom actions.
     *
     * @return self
     */
    public function setAtLibreGraphPermissionsActions($at_libre_graph_permissions_actions)
    {
        if (is_null($at_libre_graph_permissions_actions)) {
            throw new \InvalidArgumentException('non-nullable at_libre_graph_permissions_actions cannot be null');
        }
        $this->container['at_libre_graph_permissions_actions'] = $at_libre_graph_permissions_actions;

        return $this;
    }

    /**
     * Gets at_ui_hidden
     *
     * @return bool|null
     */
    public function getAtUiHidden()
    {
        return $this->container['at_ui_hidden'];
    }

    /**
     * Sets at_ui_hidden
     *
     * @param bool|null $at_ui_hidden Properties or facets (see UI.Facet) annotated with this term will not be rendered if the annotation evaluates to true. Users can set this to hide permissons.
     *
     * @return self
     */
    public function setAtUiHidden($at_ui_hidden)
    {
        if (is_null($at_ui_hidden)) {
            throw new \InvalidArgumentException('non-nullable at_ui_hidden cannot be null');
        }
        $this->container['at_ui_hidden'] = $at_ui_hidden;

        return $this;
    }
```
I strongly suggest we use those annotations for our deviations from the ms graph api

butonic:
Note that I dropped the shareID from the itemReference as it is AFAICT unused: https://github.com/owncloud/libre-graph-api/pull/112/files#diff-ef34f08de0f08ec7094fc1391308521fae6824a2ad2e5b279673c98da8a60e33L2488-R3276

butonic:
regarding the c++ build failure: I added  `#/components/schemas/sharingLinkType` which is a string with an enum and a long description:
```
    sharingLinkType:      
        type: string
        enum: [view, upload, edit, createOnly, blocksDownload]
        description: |
          The type of the link created.

          | Value          | Display name      | Description                                                     |
          | -------------- | ----------------- | --------------------------------------------------------------- |
          | view           | View              | Creates a read-only link to the driveItem.                      |
          | upload         | Upload            | Creates a read-write link to the folder driveItem.              |
          | edit           | Edit              | Creates a read-write link to the driveItem.                     |
          | createOnly     | File Drop         | Creates an upload-only link to the folder driveItem.            |
          | blocksDownload | Secure View       | Creates a read-only link that blocks download to the driveItem. |
```

My guess is that maybe the code generation does not yet handle the case that   `$ref: '#/components/schemas/sharingLinkType'` is just a `type: string` ... not an `object`? all other cshemas are objects ...

TheOneRing:
I guess we need to patch our fork of the mustache files there are 3 to 4 errors in OAISharingLinkType.cpp

butonic:
rebased on main which now has the c++ fixes, @micbar ready to merge IMO

dragotin:
Well, maybe the type just has to be changed... In XML, an enum is also not a string.

butonic:
@dragotin the codegen was already fixed in https://github.com/owncloud/libre-graph-api/pull/118

micbar (approved):
looking forward to use it 😄

PR #113, opened 2023-08-29
Add libre-graph-api-php

phil-davis:
Add generated PHP for the libre-graph-api.

I used `php-parallel-lint` to verify that the generated PHP is valid syntax. The pipelines for other languages do similar linting. I suppose that is all that we can really do to get a "warm and fuzzy" that something was generated.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -23,6 +23,11 @@ config = {
 			'repo-slug': "libre-graph-api-cpp-qt-client",
 			'branch': 'main',
 		},
+		'php': {
+			'src': "out-php",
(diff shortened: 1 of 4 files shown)


phil-davis:
This PR triggered a drone CI run which worked fine. but has a result for `continuous-integration/drone/pr`

Is there some different thing that I need to do to make the result be for the required `continuous-integration/drone/push` ?

phil-davis:
@micbar please review (and merge hopefully) or let me know who should review and approve this.

individual-it (approved):

PR #114, opened 2023-09-06
Allow using a different generator version for a language

phil-davis:
Use major version 7 of the generator to generate PHP, while continuing to use the existing 6.2 generator for the other languages.

We can't always easily move up to new releases of the generator across all languages at the same time. So this gives the flexibility to define a generator version for each language, if needed.

We use major version 7 for PHP because we are just getting started with the generated PHP, so we might as well use the latest release and learn to work with that (or fix that!)

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -27,6 +27,7 @@ config = {
 			'src': "out-php",
 			'repo-slug': "libre-graph-api-php",
 			'branch': 'main',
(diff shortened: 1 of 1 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/libre-graph-api/429/5/3
The specific lint for v7 of the generator passed in the pipeline for PHP - good.

https://drone.owncloud.com/owncloud/libre-graph-api/429/5/5
PHP  code was generated and passed a PHP lint - good.

individual-it (approved):

micbar (approved):

PR #115, opened 2023-09-06
convert data to string before giving it to preg_match

individual-it:
Add a template to fix an invalid PHP call (preg_match receives a DateTime object) see details in https://github.com/OpenAPITools/openapi-generator/pull/16513 
This fix is only temporary and should be removed when the next version of the generator is released


micbar:
> This fix is only temporary and should be removed when the next version of the generator is released

Can we create a ticket to remove it later please?

LGTM

individual-it:
@micbar done already in #116

phil-davis (approved):

PR #119, opened 2023-10-13
Add clang format

TheOneRing:


.drone.star
```diff
@@ -228,10 +229,18 @@ def generate(ctx, lang):
 def validate(lang):
 	steps = {
 		"cpp-qt-client": [
+			{
+				"name": "validate-cpp-format",
+				"image": "xianpengshen/clang-tools:17",
```
TheOneRing, line 234: There appears to be no decent docker image to be around.
I'd suggest that we push https://github.com/owncloud-ci/git-clang-format-lint to the registry.

PR #121, opened 2023-10-16
Share mounting example

butonic:
I added some missing endpoints that will be necessary to
- list sharedWithMe including permissions objects `/me/drive/sharedWithMe?$expand=permissions`
- added examples how mounted / unmounted driveItems are represented in the  `/me/drive/sharedWithMe` response
- added an example how to `POST /drives/{drive-id}/items/{parent-item-id}/children` to mount a share to the share jail, aka accepting a share. Rejecting would be done with a DELETE on the '/drives/{drive-id}/items/{item-id}/permissions/{perm-id}' but that is already specced.

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -275,6 +275,169 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
+  '/drives/{drive-id}/items/{item-id}':
+    get:
+      tags:
+        - driveItems
+      summary: Get a driveItem resource
+      operationId: GetItem
+      description: |
+        Retrieve the metadata for a driveItem in a drive. `item-id` is the ID of a driveItem.
+      parameters:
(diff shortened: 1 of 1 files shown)


api/openapi-spec/v1.0.yaml
```diff
@@ -1348,10 +1502,82 @@ paths:
...
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
                         
+  /me/drive/sharedWithMe?$expand=permissions:
+    get:
+      tags:
+        - me
+      summary: Get a list of driveItem objects and permissions shared with the current user.
+      operationId: ListSharedWithMe
```
rhafer, line 1536: The generator doesn't like the duplicated `operationId`. But AFAIK the $expand should just be defined as parameter on the original `/me/drive/sharedWithMe` specification. It might get tricky to define a proper example then, though.

rhafer (changes requested):

PR #123, opened 2023-10-16
Add location facet

dschmidt:
See https://learn.microsoft.com/en-us/graph/api/resources/geocoordinates?view=graph-rest-1.0 and https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0 🤷🏻 

Tika extracts these properties at least from images.


butonic:
nice!

butonic (approved):

micbar (approved):

PR #125, opened 2023-10-18
Switch go to openapi-generator 7

dschmidt:
We stayed on 6.6.0 to avoid a breaking change, but apparently the breaking change (Api -> API in generated struct names) seems to have been introduced before 7 already. So after a quick discussion we decided to accept this breaking change.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -18,19 +18,16 @@ config = {
 			'src': "out-typescript-axios",
 			'repo-slug': "libre-graph-api-typescript-axios",
 			'branch': 'main',
-			'generator-args': "",
 		},
 		'cpp-qt-client': {
 			'src': "out-cpp-qt-client",
 			'repo-slug': "libre-graph-api-cpp-qt-client",
 			'branch': 'main',
-			'generator-args': "",
 		},
 		'php': {
 			'src': "out-php",
 			'repo-slug': "libre-graph-api-php",
 			'branch': 'main',
-			'generator-args': "",
 		},
 	},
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -12,7 +12,6 @@ config = {
 			'src': "out-go",
 			'repo-slug': "libre-graph-api-go",
 			'branch': 'main',
-			'openapi-generator-image': 'openapitools/openapi-generator-cli:v6.6.0@sha256:08088a4625ebb8744b5cce414fe91396dbf8082e19aa32a44addcffc413dabde',
```
JanAckermann, line 15: We don't want to pin the newer version ?

dschmidt: We do, we pin the version used for all languages. This was just an exception for go to avoid a breaking change, but it slipped in anyway. We discussed it and we accept the breaking change to stay on the current version.

dschmidt:
nice, thanks a lot @rhafer

.drone.star
```diff
@@ -12,22 +12,25 @@ config = {
...
+			'generator-args': "--api-name-suffix Api",
 		},
 		'typescript-axios': {
 			'src': "out-typescript-axios",
 			'repo-slug': "libre-graph-api-typescript-axios",
 			'branch': 'main',
+			'generator-args': "",
```
dschmidt, line 21: Can't we implement a fallback if it's undefined? Don't particularly like having to add empty props three times

dschmidt: pushed

rhafer (approved):

PR #126, opened 2023-10-19
Add thumbnails relationship to driveItem

dschmidt:
See
https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0
https://learn.microsoft.com/en-us/graph/api/resources/thumbnailset?view=graph-rest-1.0
https://learn.microsoft.com/en-us/graph/api/resources/thumbnail?view=graph-rest-1.0

Tbh I'm not sure about the `thumbnailSet` definition, while I copied the exact definition from the `thumbnailSet` documentation - this page https://learn.microsoft.com/en-us/graph/api/driveitem-list-thumbnails?view=graph-rest-1.0&tabs=http#size-options lists more options and they are partially contradicting (at least the large thumbnail is said to be 1920x1920 or 800 longest on the other page)

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -3347,6 +3347,11 @@ components:
         location:
           description: Location metadata, if the item has location data. Read-only.
           $ref: '#/components/schemas/geoCoordinates'
+        thumbnails:
+          description: Collection containing ThumbnailSet objects associated with the item. Read-only. Nullable.
+          type: array
+          items:
+            $ref: '#/components/schemas/thumbnailSet'
         root:
           $ref: '#/components/schemas/root'
         trash:
@@ -3406,7 +3411,7 @@ components:
           type: string
           description: Optional. A String with format of yyyy-MM-ddTHH:mm:ssZ of DateTime indicates the expiration time of the permission.
           format: date-time
-        password: 
+        password:
           type: string
           description: Optional.The password of the sharing link that is set by the creator.
         displayName:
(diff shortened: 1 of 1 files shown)


api/openapi-spec/v1.0.yaml
```diff
@@ -3347,6 +3347,11 @@ components:
         location:
           description: Location metadata, if the item has location data. Read-only.
           $ref: '#/components/schemas/geoCoordinates'
+        thumbnails:
+          description: Collection containing ThumbnailSet objects associated with the item. Read-only. Nullable.
+          type: array
```
dschmidt, line 3352: Yes, this is an array - there can be multiple thumbnailSets for a single file - think of front- and back-artwork for music albums e.g.

dschmidt:
Thoughts on this one?

micbar:
The idea in general is a good one.

I would like to collect requirements on a broader scope for that topic.

My questions are:

- How could we implement that seriuosly? (postprocessing step writes thumbnails and stores info in file metadata)?
- How could existing clients make use of that feature?

dschmidt:
I would like to follow up on this:

- for starters I would implement it just like audio metadata. Extract it in the search service, write it into arbitrary file metadata
  * only write the original size into the metadata, we can scale to any size anyhow 
- web can make use of it by checking whether a thumbnail is available and only then make a request for the thumbnail image (saves some inexpensive requests and avoids error requests in the browser console)

dschmidt:
Any chance to make progress here any time soon?

kulmann (approved):
Looks good from my point of view...

kulmann:
@butonic @micbar I know that this doesn't have high prio at the moment... but since it can be useful for web and it's already defined in ms graph I don't see an issue bringing it to the libre graph api as well. What do you think?

dschmidt:
ping @butonic @micbar

butonic:
The idea is to save web requests if thumbnails cannot be generated. Web will only request thumbnails if the driveItem contains a thumbnail. On the server side the thumbnail metadata is created asynchronously: if we can create a thumbnail we persist the sizes in the file metadata so we can efficiently list files.
This can be done by the search service, which already reads all file to extract metadata for the audio, video and photo facets. 

The thumbnailer could also be put into the postprocessing to precreate thumbnails, but that is a different topic.

In any case the thumbnail metadata is stored in the file metadata.

We can also emit an event when the postprocessing has created the thumbnail. Then web can just wait for a thumbnail ready event.

For now this is an investment into the future where we want to use the graph api to navigate the tree.

PR #128, opened 2023-11-06
fix copy paste mistake

individual-it:
I think this is a simple copy-paste mistake and I hope I found the right description

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -837,7 +837,7 @@ paths:
     delete:
       tags:
         - drives.permissions
-      summary: Delete a share to a file or folder
+      summary: Remove access to a DriveItem
       operationId: DeletePermission
       description: |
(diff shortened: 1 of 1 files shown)


api/openapi-spec/v1.0.yaml
```diff
@@ -837,7 +837,7 @@ paths:
     delete:
       tags:
         - drives.permissions
-      summary: Delete entity from groups
+      summary: Delete a share to a file or folder
```
rhafer, line 840: > I hope I found the right description

I think the right summary would just be `Remove access to a DriveItem`. Basically the headline of the description (see line 843)

rhafer (changes requested):

api/openapi-spec/v1.0.yaml
individual-it: done

rhafer (approved):

PR #129, opened 2023-11-06
Remove version from base URLs make sharing NG endpoints beta

rhafer:
This allows us to add the version to the 'path' components so that we can make specific paths only available for a certain version.

As agreeed, this is also moving the sharing and role management related endpoints to the v1beta1 version.

Note: This might require client side changes I haven't checked that yet.


butonic (approved):
Makes sense to me ... but I don't know the fallout for generated code ... should not be an issue in ocis as we parse the urls ourself. But I don't know about the code generated for the desktop or PHP.

fschade (approved):

PR #132, opened 2023-11-17
remove type of the innererror

individual-it:
With the type set I get a problem with php(next-gen) code, because the deserelizer returns an array, but then expects an object as the error.
Do we strictly need the type here?


individual-it:
Maybe the issue is actually in the generator, see https://github.com/OpenAPITools/openapi-generator/pull/17118

PR #133, opened 2023-11-17
move to php-nextgen

individual-it:
1. its stronger typed
2. uses enums for types. I actually think the old php generator is broken for enums. It creates a class for that, but has no way to set a value of the type see https://github.com/owncloud/libre-graph-api-php/blob/main/lib/Model/SharingLinkType.php . So rather than fixing the old php, we can move on to the nextgen.


phil-davis (approved):
seems reasonable. We will see the benefits (and/or new issues) when we use it.

Issue #136, opened 2023-11-20
editing password on permission is impossible


individual-it:
Neither `Permission` nor `SharingLink` has a password property, there is no need to be able to read the password, but there should be a way to update/change a password for a link.

`Permission` only has `hasPassword` so e.g. in the PHP code we can use
`public function setHasPassword(?bool $has_password): static`
but there is no way to set a new password

micbar:
Checked the behavior of OneDrive:

1.  They are using `POST https://graph.microsoft.com/beta/drives/78186EAC57577F04/items/78186EAC57577F04!308/createLink` again and set the password there
2. They only have one link per resource.

micbar:
@butonic @individual-it 

I think we could add a `/v1beta1/drives/{drive-id}/items/{item-id}/permissions/{perm-id}/password/`

PR #137, opened 2023-11-20
feat(api): Add missing setPassword endpoint to sharing NG

micbar:
# Description

In ocis we allow more than one link per resource. As a consequence we need a dedicated call to set a  password on an existing link.

Relates to https://github.com/owncloud/libre-graph-api/issues/136

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -4040,7 +4040,7 @@ components:
       properties:
         password:
           type: string
-          description: Password. It may reuquire a password policy.
+          description: Password. It may require a password policy.
     audio:
       type: object
       description: |
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -902,6 +902,88 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
+  '/v1beta1/drives/{drive-id}/items/{item-id}/permissions/{perm-id}/setPassword':
+    post:
+      tags:
+        - drives.permissions
+      summary: Set sharing link password
+      operationId: SetPermissionPassword
+      description: |
+        Set the password of a sharing permission.
+
+        Only the `password` property can be modified this way.
+      parameters:
+        - name: drive-id
+          in: path
+          description: 'key: id of drive'
+          required: true
+          schema:
+            type: string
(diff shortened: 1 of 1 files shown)


individual-it (approved):

butonic:
no ... we need to send a PATCH request to the `permisison`.
1. the `permission` object should be able to carry an `invitation`:
```
       #invitation:
        #  description: |
        #    Details of any associated sharing invitation for this permission. For now only used when creating shares with
        #    internal users or groups. An invitation is converted into a grant and will not be part of responses.
        #  $ref: '#/components/schemas/driveItemInvite'
```
Also see https://learn.microsoft.com/en-us/graph/api/driveitem-invite?view=graph-rest-1.0&tabs=http#http-request-1 of the properties that make up the `invitation` object. We should be able to reuse the `driveItemInvite` object and uncomment the password:
```
        # internal shares do not have a password
        #password: 
        #  type: string
        #  description: The password set on the invite by the creator. Optional .
```

I left this out until now because I did not see a reason to add it, but to set a password I'd reuse the ms schema.

micbar:
@butonic i don’t want to use the invitation. Nothing to do with the link.

like I said, ms uses the createLink POST request subsequently.  So the ms scheme is not possible for us, because we have more than one link per resource

butonic:
ok, links do not have a password property up stream. I'm fine with introducing a dedicated endpoint. hm but maybe adding a `password` property to the link is easier for the webui so they can just resend the link object. might be cleaner tu use PATCH with:
```http
PATCH /drives/{drive-id}/items/{item-id}/permissions/{perm-id}

{
  "password": "mynewpassword"
}
```

`hasPassword` only appears in responses and `password` can be used in PATCH?

micbar:
> hasPassword only appears in responses and password can be used in PATCH?

hmm. Not feeling happy. I took a look at the user entity. Setpassword is a dedicated operation.

micbar:
Having a field which is only in the response and another only in the request feels confusing.

but… open to opinions

@fschade @kulmann @TheOneRing

IljaN:
From my experience it is better from a readability/auditing standpoint to have a separate endpoint to change the password.  

Else you mix security critical code mixed with "normal" entity-update logic which results in ugly code and a ugly if-statements.

butonic:
go ahead with this pr

butonic (approved):

PR #147, opened 2023-11-29
return 201 on successfull invite

individual-it:
currently ocis returns 201 and not 200 when the invite is successfull, I think that is correct so adjust the API definition

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -471,7 +471,7 @@ paths:
                     - objectId: "4c510ada-c86b-4815-8820-42cdf82c3d51"
                     '@libre.graph.permissions.actions': [ "libre.graph/driveItem/basic/read", "libre.graph/driveItem/path/update" ]
       responses:
-        '200':
+        '201':
           description: Response
           content:
             application/json:


rhafer:
MS says it's a 200: https://learn.microsoft.com/en-us/graph/api/driveitem-invite?view=graph-rest-1.0&tabs=http

I am not really fluent in HTTP, but IIUC  https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.1 says a 201 Response should come with `Location` Header (AFAIK we don't set one). The body seems to be kind of optional. AFAICT the current behavior our implementation matches more what is expected for a 200 status code. (So maybe we should fix the implemenation and return a 200)

I am pretty sure we're doing this inconsistent in a few other places as well. :cry: 

@micbar @fschade comments?

individual-it:
@rhafer to change it to 200, here a PR to ocis: https://github.com/owncloud/ocis/pull/7851

butonic:
The ms graph api returns a 200: https://learn.microsoft.com/en-us/graph/api/driveitem-invite?view=graph-rest-1.0&tabs=http#http-request-1

AFAICT the /invite endpoint is an [action](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_Actions)

From the odata spec: https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_CommonResponseStatusCodes

> 9.1.1 [Response Code 200 OK ](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode200OK) 
> 
> A request that does not create a resource returns 200 OK if it is completed successfully and the value of the resource is not null. In this case, the response body MUST contain the value of the resource specified in the request URL.
9.1.2 [Response Code 201 Created](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode201Created)
> 
> A [Create Entity](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_CreateanEntity), [Create Media Entity](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_CreateaMediaEntity), or [Invoke Action](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_InvokinganAction) request that successfully creates a resource returns 201 Created. In this case, the response body MUST contain the resource created.

Which would mean a 201.

But ... to stay compatdible with the ms graph api I would prefer to return 200.

Oh wait. Technically, the invite does not create a driveItem. It only changes a property on the drive item: permissions only exist as a property of driveItems. So in this case 200 is perfectly correct.

PR #151, opened 2023-12-06
feat: add video facet

dschmidt:
See https://learn.microsoft.com/en-us/graph/api/resources/video?view=graph-rest-1.0 and https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0 🤷


kulmann (approved):
💪

butonic (approved):

PR #154, opened 2023-12-21
add partial sharing response example and schema

butonic:
We should have a way to return errors for partially successful sharing requests.

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -599,8 +599,30 @@ paths:
                             displayName: "Albert Einstein"
                       '@libre.graph.permissions.actions': [ "libre.graph/driveItem/basic/read", "libre.graph/driveItem/path/update" ]
         '207':
-        # FIXME describe partial success response
-          description: Partial success response TODO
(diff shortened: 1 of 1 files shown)


api/openapi-spec/v1.0.yaml
```diff
@@ -599,8 +599,30 @@ paths:
...
+                value:
+                  - id: "81d5bad3-3eff-410a-a2ea-eda2d14d4474"
+                    roles: [ "7ccc2a61-9615-4063-a80a-eb7cd8e59d8" ]
+                    grantedToV2:
+                      - user:
+                          id: "4c510ada-c86b-4815-8820-42cdf82c3d51"
+                          displayName: "Albert Einstein"
+                    error:
+                      - code: notAllowed
+                        message: Account verification needed to unblock sending emails.
```
butonic, line 617: it is interesting to see that ms graph has two more properties:
one that contains a localized message that can be shown to the end user:
```json
"localizedMessage": "Kontobestätigung erforderlich, um das Senden von E-Mails zu entsperren.",
```
and one that I assume is intendet to educate the end user.
```
"fixItUrl":"http://g.live.com/8SESkydrive/VerifyAccount",
```

PR #155, opened 2023-12-22
fix: remove multi invitation example

fschade:


butonic:
Well the question is if we want to kill the example or add the proper response example: https://github.com/owncloud/libre-graph-api/pull/154

fschade:
> Well the question is if we want to kill the example or add the proper response example: #154

i'm fine with both

PR #156, opened 2023-12-29
enhancement: add permission client sync prop

fschade:
extend permission to contain the @Client.Synchronize prop

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -4091,6 +4091,9 @@ components:
           type: array
           items:
             type: string
+        '@Client.Synchronize':
+            description: Indicates if the item is synchronized with the underlying storage provider. Read-only.
+            type: boolean
         # follows the SAP UI vocabulary: https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/UI.md
         '@UI.Hidden':
           description: Properties or facets (see UI.Facet) annotated with this term will not be rendered if the annotation evaluates to true. Users can set this to hide permissons.
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -3916,6 +3916,13 @@ components:
           type: string
           description: URL that displays the resource in the browser. Read-only.
           title: shared
+        '@client.synchronize':
+            description: Indicates if the item is synchronized with the underlying storage provider. Read-only.
+            type: boolean
+          # follows the SAP UI vocabulary: https://github.com/SAP/odata-vocabularies/blob/main/vocabularies/UI.md
+        '@ui.hidden':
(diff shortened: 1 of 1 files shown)


kobergj (approved):

butonic:
This feels redundant. Clients need to safeguard against empty ids anyway. Adding a property only adds conditions clints have to check.

When we change the sharedWithMe endpoint to always contain driveItems with a remoteItem we have to invent a non existing driveItem for an unmounted / unsynced remoteItem. OneDrive actually seems to create an id for those but it has no property to distinguish if a file should be synced or not. Onedrive does not have spaces. It only synchronizes the personal drive. Shares can be synced by adding 'shortcuts' to the shared file in the personal space. In a way the onedrive personal drive is our oc10 user storage with mountpoints. you can even add multiple 'shortcuts' aka mountpoints to the personal drive and move them ito any place you like.

Technically, this is easily possible to do. 'shortcuts' in the personal drive would then implicitly cause a share to be synced. Might be easier to understand than having to manage the synced status of shares ... but collecting them in a dedicated share jail is less confusing.

We can always add 'shortcuts' later.

Back to the issue at hand: using a dedicated property to indicate the synced status. I don't like adding redundancy, especially if it allows more contradictions that we need to test and prevent. What if `id` is empty but `@Client.Synchronize` is `true` or `id` is set but `@Client.Synchronize` is `false`?

I'd not add this property and rather improve the description of the driveItem `id` te explain that it might be empty in the sharedWithMe endpoint ... however ... an empty `id` violates the identifier nature for `driveItems`: `"id": "string (identifier)",` ... 

urgh ... this sucks .... but we cannot really invent a stable idyentifier for non existing drive items. so leave it empty. and document that.

fschade:
as discussed today, '@client.synchronize' and '@ui.hidden' is now part of the remote item

butonic (approved):

rhafer:
> as discussed today, '@client.synchronize' and '@ui.hidden' is now part of the remote item

I would really love to understand why it was changed that way. Would be really cool if the commit message of the commit making that change contained some reasoning for that.

It is at least conflicting with the idea for grouping multiple share for a single use (e.g. direct share + group share) in a single `remoteItem` as outlined here: https://github.com/owncloud/ocis/pull/8016#issuecomment-1873973545

micbar:
I also don't get that change honestly. When you have more than one share per remote item the hidden and synced states are conflicting.

fschade:
@micbar i agree, that was one of my concerns too, but to be honest, please ask @butonic for details, we discussed it longer and he gave a pretty good explanation why it should be that way...

i just created the pr to be able to continue on the sharedWithMe PR.

butonic:
I remember now. @kulmann @fschade and I decided that the ui property made more sense on the remote item because users do not care about individual permissions. They only care about resources. They don't want to reject permissions. Only resources. Brb

PR #158, opened 2024-01-03
remove unused schemas

butonic:
the openapi validator [spits warnings](https://drone.owncloud.com/owncloud/libre-graph-api/528/1/2): 

```
docker.io/openapitools/openapi-generator-cli@sha256:1894bae95de139bd81b6fc2ba8d2e423a2bf1b0266518d175bd26218fe42a89b: Pulling from openapitools/openapi-generator-cli
Digest: sha256:1894bae95de139bd81b6fc2ba8d2e423a2bf1b0266518d175bd26218fe42a89b
Status: Downloaded newer image for openapitools/openapi-generator-cli@sha256:1894bae95de139bd81b6fc2ba8d2e423a2bf1b0266518d175bd26218fe42a89b
+ /usr/local/bin/docker-entrypoint.sh validate -i api/openapi-spec/v1.0.yaml
Validating spec (api/openapi-spec/v1.0.yaml)
Warnings: 
	- Unused model: directoryObject
	- Unused model: educationOrganization
	- Unused model: entity

[info] Spec has 3 recommendation(s).
```

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -3008,15 +3008,6 @@ paths:
           $ref: '#/components/responses/error'
 components:
   schemas:
-    entity:
-      description: Represents an entity.
-      readOnly: true
-      type: object
-      properties:
-        id:
-          type: string
-          description: The unique idenfier for an entity. Read-only.
-          readOnly: true
     tagAssignment:
       type: object
       required:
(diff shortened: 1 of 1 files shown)


kulmann (commented):
hm, grey area... removing the schemas would need a deprecation first and then a deletion would cause a new major version, i.e. `2.0.0`, right?!

butonic:
I don't think so. They are completely unused. Maybe @individual-it can see better if they can be found in the generated code.

rhafer:
> I don't think so. They are completely unused. Maybe @individual-it can see better if they can be found in the generated code.

The models are of course generated in the code. See: https://github.com/owncloud/libre-graph-api-go/blob/main/model_education_organization.go or https://github.com/owncloud/libre-graph-api-php/blob/main/src/Model/EducationOrganization.php, but as @butonic mentions there is no real use for them.

rhafer (approved):

PR #161, opened 2024-01-29
Adapt shareWithMe specifcation to be able to group shares by shared resource

rhafer:
- Move the `UI.Hidden` and `client.synchronize` flags to the outer drive item
- add `invitation` property to `permissions`  (to expose the share creator)
- remove `shared` property from `remoteItem`. With the share creator information move to `permssion` this can go away.
- adjust examples accordingly


rhafer:
@individual-it We needed to adapt the sharedWithMe once more :speak_no_evil:

butonic (approved):

PR #165, opened 2024-04-03
Add shortcut paths for drive root item

rhafer:
This adds a couple of new enpdoints for being able to access a drive's root more easily.

Related: https://github.com/owncloud/ocis/issues/8351

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -413,7 +413,664 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
-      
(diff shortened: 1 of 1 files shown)


rhafer:
@micbar This is duplicating lots of code from other endpoints. Any idea howto spec e.g. `drives/{driveid}/root` with so much duplication.

This is becoming unmaintainable.

micbar:
> @micbar This is duplicating lots of code from other endpoints. Any idea howto spec e.g. drives/{driveid}/root without so much duplication?

I tried composition, the generator was not ready, but we can try again.

Out of scope of this PR IMHO

PR #174, opened 2024-04-22
Update constant permission role id

amrita-shrestha:
Currently, permission role id is constant. This PR updates permission role-id.


rhafer (approved):
I'm fine merging this. But you shouldn't ever rely on any uuid mentioned in this document being used by the real-world implementation. (unless specifically mentioned)

PR #180, opened 2024-05-24
Add Basic auth to the supported security schemes

rhafer:
We still support basic auth and without this in the spec the generated go client will not include support for it.


butonic (approved):

butonic:
Hm. Where can we comment that basic auth is possible but discouraged?

rhafer:
> Hm. Where can we comment that basic auth is possible but discouraged?

Good point. Seems that `securitySchemes` allows a `description` I'll add something there ...

PR #182, opened 2024-05-28
Fix request body for UpdateDrive

rhafer:
The UpdateDrive request does not require the 'name' property. It is only required for creating a drive. We fixed this by introducing a new type 'driveUpdate' with 'name' as an optional property and deriving the 'drive' type from it with 'name' as a required property.

This is needed because the latest release of the code generator for go now does some basic validation for required attributes.

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -306,7 +306,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/drive'
+              $ref: '#/components/schemas/driveUpdate'
             examples:
               change quota:
                 value:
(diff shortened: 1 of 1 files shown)


rhafer:
Note: The change itself if not really backwards compatible. Apart from the new type. The signature of the `updateDrive` method is changing. How are we dealing with that?

I also tried the very minimal fix and just made the `name` property optional for `drive`. It would also required code changes, at least for everyone using the go module. (Drive.Name would change from `string` to `*string`)

individual-it:
@amrita-shrestha can you please check if this change will affect the PHP-SDK and the moodle extension

rhafer:
@individual-it  @amrita-shrestha  BTW, this is the diff for the generated php client: https://drone.owncloud.com/owncloud/libre-graph-api/582/5/6

PR #184, opened 2024-06-18
enhancement(api): add activities-api

fschade:


kobergj (approved):

api/openapi-spec/v1.0.yaml
```diff
@@ -4343,6 +4385,37 @@ components:
...
+        template:
+          type: object
+          required:
+            - message
+          properties:
+            message:
+              type: string
+              description: Activity description.
+            variables:
+              type: object
```
micbar, line 4417: don't we need a type here?

kobergj: No. This is just a `map[string]interface{}` in go. The interfaces can vary depending on the activity.

micbar: Hm. I am not happy, this is too vague for an API spec. Isn't that just polymorphism and we would need something like a `oneOf` logic here?

@rhafer any opinions?

fschade: oneOf, anyOf, allOf, not are all operators that expect a defined set of available items to pick from,
in the context we try to use it, that set is defined by the dynamic variables (`map[string]interface{}`) provided by the backend. In short, the values are not known before, and even not always the same (this depends on the individual activity).

in my opinion thats not a (static) polymorphism since we do not know which fields (variables) are provided by the activity, we only know which fields are needed by the template.

so it would be pretty hard to (statically) define the available set of variables before, because they can grow, are different for each activity (maybe for a subset too!?).

what we can do instead (with the loss of flexibility) define all available keywords (variables) and say thats it for the moment.

but i think thats not needed (personal opinion) since there is a close relationship between variables -> and -> the template.

the consumer than knows, i has a set of N KV pairs to hydrate the given template, there is no hard rule that the consuming part HAS to use those variables.

kobergj: > the consumer than knows, i has a set of N KV pairs to hydrate the given template, there is no hard rule that the consuming part HAS to use those variables.

Additionally we were thinking about adding the message already hydrated, but since web is not using it we kept that out for the moment.

micbar: `map[string]interface{})` is only a golang expression. Our API spec needs to work for a lot of different programing languages.

API and flexibility are two concepts which do not play well together 😄

micbar: Maybe we can use https://swagger.io/docs/specification/data-models/dictionaries/

and `additionalProperties` ?

kobergj: > map[string]interface{}) is only a golang expression. Our API spec needs to work for a lot of different programing languages.

Then don't call it `map[string]interface{}`. :smile:  Swagger calls it `Free-Form Object` https://swagger.io/docs/specification/data-models/data-types/ (they dont have ancors)

In a nutshell:

A free-form object (arbitrary property/value pairs) is defined as:
```yml
type: object
```
This is equivalent to
```yml
type: object
additionalProperties: true
```
and
```yml
type: object
additionalProperties: {}
```

micbar: Thanks for the pointer. Understanding it now. I think one more example would be good.

kobergj: :+1: There are examples for this in lines 135-141. Should we copy them here too?

micbar: ok.

PR #188, opened 2024-08-06
chore: fix typos in comments, descriptions and summaries

phil-davis:
This is just a tidy up of things that, IMO, have no syntactic impact (i.e. won't break backward-compatibility of any use of this API.


micbar (approved):
Thanks!

phil-davis:
This stuff doesn't have an real effect on the language-specific generated code, such as https://github.com/owncloud/libre-graph-api-php - so no need to regenerate that just now.

PR #189, opened 2024-08-06
chore: fix primaryRole spelling in examples

phil-davis:
One of the education user properties is `primaryRole` (not `primariyRole`)


micbar (approved):

phil-davis:
These examples don't seem to end up in the language-specific generated code, such as https://github.com/owncloud/libre-graph-api-php

So there is no need to re-generated any of that.

PR #212, opened 2026-06-14
fix(spec): ListPermissionRoleDefinitions returns an array of roleDefinitions

### LINKED ISSUE #10: ListPermissionRoleDefinitions returns *UnifiedRoleDefinition but the endpoint responds with a JSON array
## Summary

`RoleManagementApi.ListPermissionRoleDefinitions(ctx).Execute()` is typed to return a single `*UnifiedRoleDefinition` and decodes the entire response body into it — but `GET /v1beta1/roleManagement/permissions/roleDefinitions` returns a **JSON array** of role definitions. Decoding therefore fails and the method is unusable against a live oCIS.

## Details

SDK (`main`, pseudo-version `v1.0.5-0.20260528080633-1156304e59ee`), `api_role_management.go`:

```go
func (r ApiListPermissionRoleDefinitionsRequest) Execute() (*UnifiedRoleDefinition, *http.Response, error)
// ...
localVarReturnValue *UnifiedRoleDefinition           // single object
localVarPath := localBasePath + "/v1beta1/roleManagement/permissions/roleDefinitions"
```

Live oCIS 8.0.1 response for that path (a bare array):

```json
[{"@libre.graph.weight":0,"description":"View and download.", "...": "..."}, {"...": "..."}]
```

`json.Unmarshal` of a JSON array into a `*UnifiedRoleDefinition` struct returns an error, so `Execute()` always fails for this endpoint.

Note: the singular GET (`/roleDefinitions/{role-id}` → `GetPermissionRoleDefinition`) correctly returns one object; it's specifically the **list** operation that's mistyped.

## Impact

Consumers can't use the SDK to list sharing-role definitions and must hand-roll this one endpoint — e.g. owncloud/ocis-mcp-server#24 / owncloud/ocis-mcp-server#13 keep `get_sharing_roles` on a raw HTTP call for exactly this reason.

## Suggested fix

The list operation's response schema should be an array (`[]UnifiedRoleDefinition`) or a collection wrapper — most likely a fix in the OpenAPI spec the SDK is generated from, then regenerate.

## Environment

- libre-graph-api-go: `main` (`v1.0.5-0.20260528080633-1156304e59ee`)
- oCIS: 8.0.1

dj4oC:
## Problem

The `GET /v1beta1/roleManagement/permissions/roleDefinitions` (`ListPermissionRoleDefinitions`) operation declares its `200` response schema as a **single** `unifiedRoleDefinition`:

```yaml
schema:
  $ref: '#/components/schemas/unifiedRoleDefinition'
example:
  value:
    - id: ...
```

But the endpoint returns a **JSON array** of role definitions (verified against a live oCIS 8.0.1 — the body starts with `[{ ... }, ...]`). Two problems follow:

1. Generated clients fail to decode the response. The Go client's `Execute()` returns `*UnifiedRoleDefinition` and json-decodes the whole array-body into a single struct, which errors — so the method is unusable (downstream has to hand-roll this endpoint; see owncloud/libre-graph-api-go#10 and owncloud/ocis-mcp-server#24).
2. The schema and the example disagree (singular object vs a `value:`-wrapped array), and both disagree with the live bare-array response.

## Fix

Make the `200` response an array and correct the example to a bare array:

```yaml
schema:
  type: array
  items:
    $ref: '#/components/schemas/unifiedRoleDefinition'
example:
  - id: ...
```

The singular `GET .../roleDefinitions/{role-id}` (`GetPermissionRoleDefinition`) is unchanged — it correctly returns one object.

## Verification

- `openapi-generator-cli v7.8.0 validate -i api/openapi-spec/v1.0.yaml` → **No validation issues detected**.
- Regenerated the Go client from this spec: `func (r ApiListPermissionRoleDefinitionsRequest) Execute() ([]UnifiedRoleDefinition, *http.Response, error)` with `localVarReturnValue []UnifiedRoleDefinition` — the list now decodes correctly.

Fixes owncloud/libre-graph-api-go#10

### DIFF
--- a/api/openapi-spec/v1.0.yaml
+++ b/api/openapi-spec/v1.0.yaml
@@ -3971,43 +3971,44 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/unifiedRoleDefinition'
+                type: array
+                items:
+                  $ref: '#/components/schemas/unifiedRoleDefinition'
               example:
-                value:
-                  - id: b1e2218d-eef8-4d4c-b82d-0f1a1b48f3b5
-                    description: "Allows reading the shared file or folder"
-                    displayName: Viewer
-                    rolePermissions:
-                      - allowedResourceActions:
-                          - libre.graph/driveItem/basic/read
-                          - libre.graph/driveItem/permissions/read
-                          - libre.graph/driveItem/upload/create
-                        condition: "exists @Resource.File"
-                    '@libre.graph.weight': 1
-                  - id: fb6c3e19-e378-47e5-b277-9732f9de6e21
-                    description: "Allows reading and writing the shared file or folder"
-                    displayName: Editor
-                    rolePermissions:
-                      - allowedResourceActions:
-                          - libre.graph/driveItem/standard/allTasks
-                        condition: "exists @Resource.File"
-                    '@libre.graph.weight': 2
-                  - id: 312c0871-5ef7-4b3a-85b6-0e4074c64049
-                    description: "Allows managing a space"
-                    displayName: Manager
-                    rolePermissions:
-                      - allowedResourceActions:
-                          - libre.graph/drive/standard/allTasks
-                        condition: "exists @Resource.File"
-                    '@libre.graph.weight': 3
-                  - id: 4916f47e-66d5-49bb-9ac9-748ad00334b
-                    description: "Allows creating new files"
-                    displayName: File Drop
-                    rolePermissions:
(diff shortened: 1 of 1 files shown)


dj4oC:
@owncloud/ocis-maintainers — friendly nudge: this PR has had no review since it was opened on 2026-06-14 and CI is green. Could someone take a look when you have a moment? Happy to rebase or adjust if anything needs changing.

kobergj (approved):

